PR #6, opened 2021-05-28
Add the option to activate silent mode for actions which disables sending of .started and .finished events

yeahservice:
* Actions can be run in silent mode, meaning no `.started/.finished` events will be sent by the job-executor-service. This
is particular useful when not matching `.triggered` events but e.g. `.finished` events where responding
with `.started/.finished` events does not make sense. To enable silent mode simply set it to true for the corresponding
action. By default silent mode is disabled for each action.

```yaml
actions:
  - name: "Run locust"
    silent: true
```

* Logs of all completed tasks are now appended to the `.finished` event, currently was only sending the output of the last task. For failed `.finished` events only the output of the errored job will be taken (same behavior as before).
* The same k8s clientset will now be used for each task, instead of creating a new one for each task.


README.md
```diff
@@ -191,9 +190,24 @@ E.g.:
...
+actions:
+  - name: "Run locust"
+    silent: true
+```
+
 ### Remote Control Plane
-If you are using the service in a remote control plane setup make sure the distributor is configured to forward all events used in 
-the `job/config.yaml`. Just edit the `PUBSUB_TOPIC` environment variable in the distributor deployment configuration to fit your needs. 
+
+If you are using the service in a remote control plane setup make sure the distributor is configured to forward all
```
didiladi, line 208: :rocket:

pkg/k8s/connect.go
```diff
@@ -1,12 +1,28 @@
...
 )
 
+// K8s is used to interact with kubernetes jobs
+type K8s struct {
+}
+
+//go:generate mockgen -source=connect.go -destination=fake/connect_mock.go -package=fake Interface
+
+// Interface is used to interact with kubernetes jobs
+type Interface interface {
```
didiladi, line 17: Can you name this to indicate that it groups methods for kubernetes? `Interface` is a bit generic. :D

didiladi: Maybe name the interface `K8s` and the struct `k8sImpl`

yeahservice: Renamed:
* package `k8s` to `k8sutils`
* `Interface` to `K8s`
* struct `K8s` to `K8sImpl`

yeahservice: Also made k8sImpl non-exported and added a function `NewK8s` that creates and returns an k8sImpl

pkg/k8sutils/connect.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	keptnv2 "github.com/keptn/go-utils/pkg/lib/v0_2_0"
+	keptnutils "github.com/keptn/kubernetes-utils/pkg"
+	"k8s.io/client-go/kubernetes"
+	"keptn-sandbox/job-executor-service/pkg/config"
+)
+
+// k8sImpl is used to interact with kubernetes jobs
+type k8sImpl struct {
```
didiladi, line 11: Maybe for a future improvement: We could store the `*kubernetes.Clientset` in `k8sImpl`. This would result in shorter methods in `K8s`.

didiladi (approved):

pkg/k8sutils/connect.go
yeahservice: Yes, some of the method signatures are getting quite long

PR #15, opened 2021-06-09
Introduce possibility to make kubernetes secrets available as environment variables

yeahservice:
## Kubernetes secrets as environment variables 

Environment variables now need to be declared with their source either from the incoming cloud event (`valueFrom: event`) or from kubernetes secrets (`valueFrom: secret`). If `valueFrom` is unknown or not set the job creation fails with an error.

The following configuration looks up a kubernetes secret with the name `locust-secret` and all key/value pairs of the secret will be available as separate environment variables in the job.

```yaml
        cmd: "locust --config /keptn/locust/locust.conf -f /keptn/locust/$(FILE) --host $(HOST)"
        env:
          - name: locust-secret
            valueFrom: secret
```

With the secret below, there will be two environment variables available in the job. `HOST` with the value `https://keptn.sh` and `FILE` with the value `basic.py`

```shell
kubectl -n keptn create secret generic locust-secret --from-literal=HOST=https://keptn.sh --from-literal=FILE=basic.py -oyaml --dry-run=client
```

```yaml
apiVersion: v1
data:
  FILE: YmFzaWMucHk=
  HOST: aHR0cHM6Ly9rZXB0bi5zaA==
kind: Secret
metadata:
  creationTimestamp: null
  name: locust-secret
  namespace: keptn
```

The `Environment` of the job pod looks like this:
```
Environment:
  HOST: <set to the key 'HOST' in secret 'locust-secret'> Optional: false
  FILE: <set to the key 'FILE' in secret 'locust-secret'> Optional: false
```

## Other changes

* Add `apiVersion` to job config and enforce supported version. Currently `v1` is supported and the loading of the configuration fails if the version does not match.
* Move `clientset` and `namespace` to `k8sImp` struct to reduce number of parameters for related `k8s` functions


pkg/k8sutils/connect.go
```diff
@@ -9,30 +9,33 @@ import (
 
 // k8sImpl is used to interact with kubernetes jobs
 type k8sImpl struct {
+	clientset kubernetes.Interface
```
didiladi, line 12: :rocket:

pkg/k8sutils/job_test.go
```diff
@@ -42,20 +46,47 @@ const testTriggeredEvent = `
...
 
-func TestPrepareJobEnv(t *testing.T) {
+func TestPrepareJobEnv_WithNoValueFrom(t *testing.T) {
 	task := config.Task{
 		Env: []config.Env{
 			{
-				Name:  "HOST",
-				Value: "$.data.deployment.deploymentURIsLocal[0]",
+				Name:  "DEPLOYMENT_STRATEGY",
+				Value: "$.data.deployment.undeploymentstrategy",
```
didiladi, line 54: That made me laugh! :smile:

didiladi (approved):
Cool, awesome change! I like the additional test coverage! :rocket: :muscle:

PR #31, opened 2021-06-25
Extend command to allow declaration of arrays and and args to be passed through to the k8s job

yeahservice:
With the current way it is not really possible to write more complex commands, since everything gets just splitted by white spaces and added as command to the k8s job, which can lead to issues with chained commands `&&`. Let users decide how exactly they want to split commands and also allow definition of args, passed through to the k8s job.

Current declaration:
```yaml
cmd: "locust --config /keptn/locust/locust.conf -f /keptn/locust/basic.py --host $(HOST)"
```

New declaration:
```yaml
cmd:
  - locust
args:
  - '--config'
  - /keptn/locust/locust.conf
  - '-f'
  - /keptn/locust/basic.py
  - '--host'
  - $(HOST)
```

Bumped config version to `v2` because it is not backwards compatible.

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -8,7 +8,7 @@ import (
 )
 
 const simpleConfig = `
-apiVersion: v1
+apiVersion: v2
 actions:
   - name: "Run locust"
     events:
@@ -22,11 +22,14 @@ actions:
           - locust/basic.py
           - locust/import.py
         image: "locustio/locust"
-        cmd: "locust -f /keptn/locust/locustfile.py"
+        cmd:
+          - locust
+          - '-f'
+          - /keptn/locust/locustfile.py
 `
 
 const complexConfig = `
-apiVersion: v1
+apiVersion: v2
 actions:
   - name: "Run locust"
     events:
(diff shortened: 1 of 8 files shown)


pkg/config/config_test.go
```diff
@@ -22,11 +22,14 @@ actions:
           - locust/basic.py
           - locust/import.py
         image: "locustio/locust"
-        cmd: "locust -f /keptn/locust/locustfile.py"
+        cmd:
+          - locust
+          - '-f'
```
didiladi, line 27: There is a `args` missing.

pkg/file/file_test.go
```diff
@@ -25,7 +25,10 @@ actions:
           - /helm/values.yaml
           - locust
         image: "locustio/locust"
-        cmd: "locust -f /keptn/locust/basic.py"
+        cmd:
+          - locust
+          - '-f'
```
didiladi, line 30: `args`?

didiladi (commented):
Just wondering: Is the following syntax possible (without the `args`)?

```
cmd:
      - locust
      - '--config'
      - /keptn/locust/locust.conf
      - '-f'
      - /keptn/locust/basic.py
      - '--host'
      - $(HOST)
```

yeahservice:
> Just wondering: Is the following syntax possible (without the args)?
It is valid and (hopefully) runs without any issues, did not test this more complex example. I mixed the examples a bit to show that off but it might be better to just use `cmd` and `args` everywhere to show off good practice and keep it consistent.

pkg/config/config_test.go
yeahservice: Moved arguments to `args`

pkg/file/file_test.go
yeahservice: Moved arguments to `args`

didiladi (approved):

PR #36, opened 2021-07-07
Add new env variable source `string`

### LINKED ISSUE #34: Support env variables from string
Currently one can use env with source either from `valueFrom: secret` or from `valueFrom: event`. It would also be useful to just define common environment variables, something like:

```yaml
- name: TF_DATA_DIR
  value: /tmp/.terraform
  valueFrom: string
```

patrickpichler:
There is now a new env variable source `string` which allows you to
specify the value of an env var, by providing the `value` property.

This fixes #34

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,34 @@ metadata:
   namespace: keptn
 ```
 
+#### From string literal
+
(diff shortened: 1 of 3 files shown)


yeahservice (approved):
lgtm

PR #38, opened 2021-07-12
Feature/37/add working dir spec

patrickpichler:
The `k8s` interface has been refactored to allow for easier testing and 
there is now the possibility to specify the CWD of the command of a task.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,29 @@ actions:
             valueFrom: event
 ```
 
+### Specifying the working directory
+
+Since all files are hosted by default under `/keptn` and some tools only operate on the current
+working directory, it is also possible to switch the working directory of the container.
+This can be achieved by setting the `workingDirectory` property in a task object.
+
+Here an example:
+```yaml
+apiVersion: v2
+actions:
+  - name: "Print files"
+    events:
+      - name: "sh.keptn.event.sample.triggered"
+    tasks:
+      - name: "Show files in bin"
+        image: "alpine"
+        workingDirectory: "/bin"
+        cmd:
+          - ls
+```
+
+In this example, the `ls` command will be run in the `/bin` folder.
+
 ### Event Matching
 
 The tasks of an action are executed if the event name matches. Wildcards can also be used, e.g.
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -38,13 +38,14 @@ type JSONPath struct {
 
 // Task this is the actual task which can be triggered within an Action
 type Task struct {
-	Name      string     `yaml:"name"`
-	Files     []string   `yaml:"files"`
-	Image     string     `yaml:"image"`
-	Cmd       []string   `yaml:"cmd"`
-	Args      []string   `yaml:"args"`
-	Env       []Env      `yaml:"env"`
-	Resources *Resources `yaml:"resources"`
+	Name       string     `yaml:"name"`
+	Files      []string   `yaml:"files"`
+	Image      string     `yaml:"image"`
+	Cmd        []string   `yaml:"cmd"`
+	Args       []string   `yaml:"args"`
+	Env        []Env      `yaml:"env"`
+	Resources  *Resources `yaml:"resources"`
+	WorkingDir string     `yaml:"workingDir"`
 }
 
 // Env value from the event which will be added as env to the job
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -61,6 +61,29 @@ actions:
             valueFrom: event
 ```
 
+### Specifying the working directory
+
+Since all files are hosted by default under `/keptn` and some tools only operate on the current
+working directory, it is also possible to switch the working directory of the container.
+This can be archieved by setting the `workingDirectory` property in a task object.
```
yeahservice, line 68:
```suggestion
This can be achieved by setting the `workingDirectory` property in a task object.
```

pkg/eventhandler/eventhandlers.go
```diff
@@ -114,14 +114,25 @@ func (eh *EventHandler) startK8sJob(k8s k8sutils.K8s, action *config.Action, jso
 		// k8s job name max length is 63 characters, with the naming scheme below up to 999 tasks per action are supported
 		jobName := "job-executor-service-job-" + eh.Event.ID()[:28] + "-" + strconv.Itoa(index+1)
 
-		jobErr := k8s.CreateK8sJob(jobName, action, task, eh.EventData, eh.JobSettings, jsonEventData)
+		jobName, err := k8s.CreateK8sJob(jobName, action, task, eh.EventData, eh.JobSettings, jsonEventData)
```
yeahservice, line 117: As we discussed already, `jobName` as return value would not be necessary here, since it is the same value as passed in?

pkg/config/config.go
```diff
@@ -38,13 +38,14 @@ type JSONPath struct {
...
-	Env       []Env      `yaml:"env"`
-	Resources *Resources `yaml:"resources"`
+	Name       string     `yaml:"name"`
+	Files      []string   `yaml:"files"`
+	Image      string     `yaml:"image"`
+	Cmd        []string   `yaml:"cmd"`
+	Args       []string   `yaml:"args"`
+	Env        []Env      `yaml:"env"`
+	Resources  *Resources `yaml:"resources"`
+	WorkingDir string     `yaml:"workingDir"`
```
yeahservice, line 48: I guess extending one of the configs in [config_test.go](https://github.com/keptn-sandbox/job-executor-service/blob/main/pkg/config/config_test.go) just to see if unmarshalling works without errors would be nice.

yeahservice (changes requested):

pkg/eventhandler/eventhandlers.go
patrickpichler: Can we be sure, that it will always be the same? 

If yes, then I will get rid of it :+1:

patrickpichler: Should be fixed :+1:

yeahservice (approved):

pkg/k8sutils/job.go
```diff
@@ -28,8 +28,10 @@ type JobSettings struct {
 	DefaultResourceRequirements                  *v1.ResourceRequirements
 }
 
-// CreateK8sJob creates a k8s job with the job-executor-service-initcontainer and the job image of the task and waits until the job finishes
-func (k8s *k8sImpl) CreateK8sJob(jobName string, action *config.Action, task config.Task, eventData *keptnv2.EventData, jobSettings JobSettings, jsonEventData interface{}) error {
+// CreateK8sJob creates a k8s job with the job-executor-service-initcontainer and the job image of the task
+// and returns the name of the job created
```
didiladi, line 32: Nitpick:

> returns the name of the job created

pkg/eventhandler/eventhandlers.go
```diff
@@ -114,14 +114,25 @@ func (eh *EventHandler) startK8sJob(k8s k8sutils.K8s, action *config.Action, jso
 		// k8s job name max length is 63 characters, with the naming scheme below up to 999 tasks per action are supported
 		jobName := "job-executor-service-job-" + eh.Event.ID()[:28] + "-" + strconv.Itoa(index+1)
 
-		jobErr := k8s.CreateK8sJob(jobName, action, task, eh.EventData, eh.JobSettings, jsonEventData)
+		err := k8s.CreateK8sJob(jobName, action, task, eh.EventData, eh.JobSettings, jsonEventData)
```
didiladi, line 117: Nice! Splitting the creation + the polling is a good idea! :muscle:

didiladi (approved):

PR #42, opened 2021-07-14
Clean up makefile, add targets for building docker images and relocate lint source directory

yeahservice:
Relocate lint source directory to `cmd` to follow go conventions.
Move makefile to root directory, clean up targets and add targets for building docker images to make @patrickpichler happy.


Makefile
```diff
@@ -0,0 +1,21 @@
+.PHONY: build-lint build-docker-service build-docker-initcontainer build-docker
+
+ifndef TAG
```
patrickpichler, line 3: You can use `TAG ?= latest` here instead of the if

patrickpichler (approved):

PR #43, opened 2021-07-14
Make poll duration for single tasks configureable

yeahservice:
The default settings allow a job to run for 5 min until the job executor service cancels the task execution. The default
value can be overwritten for each task and is declared as seconds. The setting below would result in a poll duration of
20 minutes for this specific task:

```yaml
tasks:
  - name: "Run locust smoke tests"
    ...
    maxPollDuration: 1200
```

Internally we use 5 seconds between each poll attempt. So the above setting would result in 240 poll attempts.

### DIFF
--- a/pkg/eventhandler/eventhandlers.go
+++ b/pkg/eventhandler/eventhandlers.go
@@ -6,10 +6,16 @@ import (
 	"keptn-sandbox/job-executor-service/pkg/config"
 	"keptn-sandbox/job-executor-service/pkg/k8sutils"
 	"log"
+	"math"
 	"strconv"
 
 	cloudevents "github.com/cloudevents/sdk-go/v2" // make sure to use v2 cloudevents here
 	keptnv2 "github.com/keptn/go-utils/pkg/lib/v0_2_0"
+)
+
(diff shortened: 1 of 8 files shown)


pkg/eventhandler/eventhandlers.go
```diff
@@ -131,7 +136,11 @@ func (eh *EventHandler) startK8sJob(k8s k8sutils.K8s, action *config.Action, jso
 			}
 		}()
 
-		jobErr := k8s.AwaitK8sJobDone(jobName)
+		maxPollCount := defaultMaxPollCount
+		if task.MaxPollDuration != nil {
+			maxPollCount = *task.MaxPollDuration / pollIntervalInSeconds
```
patrickpichler, line 141: I guess a `ceil` would be better than a `floor` here.

patrickpichler (approved):

pkg/eventhandler/eventhandlers.go
yeahservice: done

didiladi (approved):

PR #62, opened 2021-08-16
Prepare release 0.1.4

yeahservice:


releasenotes/releasenotes_V0.1.4.md
```diff
@@ -0,0 +1,11 @@
+# Release Notes v0.1.4
+
+Compatible with Keptn 0.8.7
+
+## New Features
```
didiladi, line 5: There also was a new feature from @thschue in the meanwhile: #52 and #57.

I think it also makes sense to wait for #63.

PR #63, opened 2021-08-16
Event data formatting

### LINKED ISSUE #59: Problem with `valueFrom: event` formatting
For some specific services you may need to pass the whole Keptn event into the container using an environment variable which can currently be done using the following syntax:

```
env:
          - name: EVENT
            value: "$"
            valueFrom: event
```

When reading through [the code](https://github.com/keptn-sandbox/job-executor-service/blob/770dc0ffb05d6cf5f32c27cda233d2556e1c6a9a/pkg/eventhandler/eventhandlers.go#L105) you get the impression that the event data is stored as JSON which would be easy to work with once read through the environment variable but it seems that the event is formatted as an [map[string]interface](https://github.com/keptn-sandbox/job-executor-service/blob/c45ce5924e559f26b422e3c8a90c37f8587ca723/pkg/eventhandler/eventhandlers.go#L70) instead which leaves you with the following value when using the environment variable:

```bash
map[data:map[configurationChange:map[values:map[image:docker.io/keptnexamples/carts:0.12.3]] deployment:map[deploymentNames:[direct] deploymentURIsLocal:[http://carts.sockshop-dev:80] deploymentURIsPublic:[http://carts.sockshop-dev.IP_ADDRESS.nip.io:80] deploymentstrategy:direct gitCommit:b02c3afdcd539dbe33c461e75d4d5365acff914b] message: project:sockshop result:pass service:carts stage:dev status:succeeded test:map[teststrategy:functional]] id:daa042d5-31b8-443e-826b-c26505da84ff shkeptncontext:fc3f12ac-cda8-483c-944a-985bfef3b102 source:shipyard-controller time:2021-08-05 17:29:51.323554006 +0000 UTC]
```

This format has proven quite hard to cast (extract the keys and values). Is there a way to get the same data in another format like JSON or YAML? Or is there a preferred way of casting the result to a Keptn event?

TannerGabriel:
Allowing the user to specify `JSON` and `YAML` as formats for the environment event data. 

```
- name: EVENT
  value: "$"
  valueFrom: event
  as: json
```

If the `as` field is left empty the event will either be passed as JSON if it is a map or as a string if it is a single value.

```
- name: EVENT
  value: "$"
  valueFrom: event
```

## Fixes
#59

### DIFF
--- a/pkg/k8sutils/job.go
+++ b/pkg/k8sutils/job.go
@@ -3,15 +3,15 @@ package k8sutils
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"github.com/PaesslerAG/jsonpath"
 	"gopkg.in/yaml.v2"
 	"keptn-sandbox/job-executor-service/pkg/config"
 	"reflect"
-	"strings"
 	"strconv"
+	"strings"
 	"time"
-	"log"
 
 	keptnv2 "github.com/keptn/go-utils/pkg/lib/v0_2_0"
 
@@ -235,7 +235,7 @@ func (k8s *k8sImpl) prepareJobEnv(task config.Task, eventData *keptnv2.EventData
 
 		switch env.ValueFrom {
 		case envValueFromEvent:
-			generatedEnv, err = generateEnvFromEvent(env, jsonEventData, env.Formatting)
+			generatedEnv, err = generateEnvFromEvent(env, jsonEventData)
 		case envValueFromSecret:
 			generatedEnv, err = k8s.generateEnvFromSecret(env)
 		case envValueFromString:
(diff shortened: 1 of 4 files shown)


pkg/k8sutils/job.go
```diff
@@ -258,13 +263,41 @@ func (k8s *k8sImpl) prepareJobEnv(task config.Task, eventData *keptnv2.EventData
...
+
+		if err != nil {
+			log.Printf("Error converting environment variable value to YAML: %e",err)
+		}
+
+		value = string(yamlString[:])
+	} else if reflect.ValueOf(value).Kind() == reflect.Map {
+		log.Println("Values is map")
+
+		jsonString, err := json.Marshal(value)
```
didiladi, line 292: Ok, so your solution is implementing option1+2 in #59. Its fine by me, since we don't parse the map putput. However, this is a breaking change which needs to be documented.

didiladi (commented):
The implementation looks good to me. :rocket: Can you please also document the new functionality + breaking change in the readme?

pkg/k8sutils/job.go
TannerGabriel: Documentation has been added. Should the breaking change also be mentioned in the upcoming release?

didiladi: Cool, thanks @TannerGabriel! :rocket: Yes, makes sense. @yeahservice can you please add a note to #62?

didiladi (approved):

yeahservice (changes requested):

pkg/k8sutils/job.go
yeahservice: Will do

pkg/k8sutils/job.go
```diff
@@ -263,13 +268,41 @@ func (k8s *k8sImpl) prepareJobEnv(task config.Task, eventData *keptnv2.EventData
...
+		value = string(jsonString[:])
+	} else if strings.EqualFold(formatting, "yaml") {
+		yamlString, err := yaml.Marshal(value)
+
+		if err != nil {
+			log.Printf("Error converting environment variable value to YAML: %e",err)
+		}
+
+		value = string(yamlString[:])
+	} else if reflect.ValueOf(value).Kind() == reflect.Map {
```
yeahservice, line 294: The logic in the map and json if block is the same, could merge it
```golang
	if strings.EqualFold(formatting, "json") || reflect.ValueOf(value).Kind() == reflect.Map {
		jsonString, err := json.Marshal(value)

		if err != nil {
			log.Printf("Error converting environment variable value to JSON: %e",err)
		}

		value = string(jsonString[:])
	}

```

yeahservice (changes requested):

pkg/k8sutils/job.go
```diff
@@ -263,13 +268,41 @@ func (k8s *k8sImpl) prepareJobEnv(task config.Task, eventData *keptnv2.EventData
 	return jobEnv, nil
 }
 
-func generateEnvFromEvent(env config.Env, jsonEventData interface{}) ([]v1.EnvVar, error) {
+func generateEnvFromEvent(env config.Env, jsonEventData interface{}, formatting string) ([]v1.EnvVar, error) {
```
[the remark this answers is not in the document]
TannerGabriel: Yes, that seems to be a mistake.

pkg/k8sutils/job.go
```diff
@@ -263,13 +268,41 @@ func (k8s *k8sImpl) prepareJobEnv(task config.Task, eventData *keptnv2.EventData
...
 	value, err := jsonpath.Get(env.Value, jsonEventData)
 	if err != nil {
 		return nil, fmt.Errorf("could not add env with name %v, value %v, valueFrom %v: %v", env.Name, env.Value, env.ValueFrom, err)
 	}
 
+	if strings.EqualFold(formatting, "json") {
+		jsonString, err := json.Marshal(value)
+
+		if err != nil {
+			log.Printf("Error converting environment variable value to JSON: %e",err)
```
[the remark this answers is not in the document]
TannerGabriel: Sounds resonable!

yeahservice (approved):

yeahservice:
lgtm

PR #64, opened 2021-08-18
feat: Add start and end event metadata for test finished events

didiladi:
This commit extends the job ececutor service to send back the start and
end event metadata on test.finished events.

Still TODO:
* Add tests
* Start a discussion if this behaviour should be added for more shipyard tasks beside `test` tasks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@
         - [Poll duration for jobs](#poll-duration-for-jobs)
         - [Job namespace](#job-namespace)
         - [Remote Control Plane](#remote-control-plane)
+        - [Additional Event Data](#additional-event-data)
     - [How to validate a job configuration](#how-to-validate-a-job-configuration)
     - [Endless Possibilities](#endless-possibilities)
     - [Credits](#credits)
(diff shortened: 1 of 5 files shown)


TannerGabriel (approved):
I just tried out your PR since I was facing an evaluation error: `evaluation.triggered event does not contain evaluation timeframe` and this seems to fix my problem. The implementation also looks good to me.

yeahservice:
> Start a discussion if this behavior should be added for more shipyard tasks beside test tasks.

I would say supporting just test events is enough for this PR and probably covers most of the use cases. The other important one would probably be deployment events with the `deploymentURIs` as data. But this would require something like getting the event data out of the job execution or from the keptn configuration files and would be dependent on what tool is used for the deployment. In particular, a more general approach for getting data out of job into finished events would be needed.

yeahservice (approved):

yeahservice:
@didiladi I added a section in the README and a test. In addition I changed the matching for adding the test data from `sh.keptn.event.test.` to only `sh.keptn.event.test.triggered`. E.g. in case someone has job configurations listening on `sh.keptn.event.test.finished` for example for notifications we don't overwrite the timestamps and mess up evaluation tasks later in the sequence.

README.md
```diff
@@ -470,6 +472,34 @@ If you are using the service in a remote control plane setup make sure the distr
...
+timestamp to do an evaluation.
+
+The job executor service currently adds the following data to specific event types:
+
+* Incoming `sh.keptn.event.test.triggered`
+    * Outgoing `sh.keptn.event.test.finished` events contain a `start` and `end` timestamp, marking the beginning and
+      the end time of the job responsible for handling the event
+      ```json
+      {
+        "type": "sh.keptn.event.test.triggered",
```
didiladi, line 488: This should be `sh.keptn.event.test.finished`.

didiladi (commented):
@yeahservice thanks for adding the documentation and the test :muscle:. Just found a minor thing in the documentation.

PR #67, opened 2021-08-18
Feature/make namespace selectable

thschue:


pkg/k8sutils/job.go
```diff
@@ -78,10 +78,13 @@ func (k8s *k8sImpl) CreateK8sJob(jobName string, action *config.Action, task con
 		return fmt.Errorf("could not prepare env for job %v: %v", jobName, err.Error())
 	}
 
+	if namespace != "" {
+		namespace = k8s.namespace
```
yeahservice, line 82: I would suggest deleting the namespace from the k8s struct and just use the namespace passed in as parameter of the function. Because with the current approach it can be error-prone to overwrite the namespace in each function call (see functions `DeleteK8sJob` and `GetLogsOfPod` where k8s.namespace is used but not overwritten yet => would probably result in an error)

The task loop can then decide if to pass in the default namespace or the task specific namespace to each function call.

yeahservice (changes requested):

pkg/config/config.go
```diff
@@ -47,6 +47,7 @@ type Task struct {
 	Resources       *Resources `yaml:"resources"`
 	WorkingDir      string     `yaml:"workingDir"`
 	MaxPollDuration *int       `yaml:"maxPollDuration"`
+	Namespace       string     `yaml:"namespace,omitempty"`
```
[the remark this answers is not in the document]
yeahservice: done

pkg/k8sutils/job.go
yeahservice: refactored

yeahservice (changes requested):
Readme update missing, otherwise good to go

yeahservice (approved):

PR #99, opened 2021-11-21
feat: Add support for ttlSecondsAfterFinished (#44)

### LINKED ISSUE #44: Add support for ttlSecondsAfterFinished to cleanup completed jobs
Over time you can end up with many Jobs in the `Completed` status. It would be handy to be able to enable optional support for [`ttlSecondsAfterFinished`](https://kubernetes.io/docs/concepts/workloads/controllers/job/#clean-up-finished-jobs-automatically) to cleanup these Jobs. This is a beta feature in Kubernetes 1.21 and alpha before that (available as an enabled alpha feature in the EKS 1.20 [release](https://aws.amazon.com/blogs/containers/amazon-eks-1-20-released/)).

This feature might be added as a controller argument to enable the feature but that is disabled by default.

TannerGabriel:
Implements the` TTLSecondsAfterFinished` parameter to cleanup Jobs after they have reached the `Completed` status.

## Usage

```
apiVersion: v2
actions:
  - name: "Print files"
    events:
      - name: "sh.keptn.event.sample.triggered"
    tasks:
      - name: "Show files in bin"
        image: "alpine"
        workingDir: "/bin"
        cmd:
          - ls
        ttlSecondsAfterFinished: 60
``` 

## Fixes
#44

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -48,6 +48,7 @@ type Task struct {
 	WorkingDir      string     `yaml:"workingDir"`
 	MaxPollDuration *int       `yaml:"maxPollDuration"`
 	Namespace       string     `yaml:"namespace"`
+	TTLSecondsAfterFinished *int32 `yaml:"ttlSecondsAfterFinished"`
 }
 
 // Env value from the event which will be added as env to the job
--- a/pkg/k8sutils/job.go
+++ b/pkg/k8sutils/job.go
@@ -79,6 +79,13 @@ func (k8s *k8sImpl) CreateK8sJob(jobName string, action *config.Action, task con
 		return fmt.Errorf("could not prepare env for job %v: %v", jobName, err.Error())
 	}
 
+	var TTLSecondsAfterFinished *int32
+	if task.TTLSecondsAfterFinished == nil {
+		*TTLSecondsAfterFinished = 600
+	} else {
+		TTLSecondsAfterFinished = task.TTLSecondsAfterFinished
+	}
+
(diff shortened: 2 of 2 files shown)


TannerGabriel:
@didiladi Should we also add a default value for the parameter when it is not set?

christian-kreuzberger-dtx:
@didiladi This looks good to me. I'm unsure about a default value, but I guess it makes sense to clean up afterwards (though I would set the default rather high, maybe an hour or so?).

@TannerGabriel Quick question: What happens when this value is used on Kubernetes 1.20 or older?

thschue (dismissed):
lgtm

didiladi:
> @didiladi This looks good to me. I'm unsure about a default value, but I guess it makes sense to clean up afterwards (though I would set the default rather high, maybe an hour or so?).

I agree with @christian-kreuzberger-dtx, a high default value makes sense. :+1: I'd would set it even higher (like 6 hours). Otherwise you might end up in the situation that you are unable to investigate why a given job failed, because the job is already gone. What do you think @christian-kreuzberger-dtx and @TannerGabriel?

christian-kreuzberger-dtx:
6h seems reasonable to me. Not too long, but long enough to investigate why something failed.
@TannerGabriel can you add the default value to your PR please?

TannerGabriel:
@christian-kreuzberger-dtx Sure, I will add the default value today.

christian-kreuzberger-dtx (approved):
LGTM

PR #112, opened 2021-12-15
chore: Introduce new CI and pre-release pipeline, level up helm charts

christian-kreuzberger-dtx:
## This PR

- introduces a re-usable unit test action
- updates CI pipeline to make use of reusable keptn/gh-automation workflows/actions
- introduces a pre-release pipeline which makes use of keptn/gh-automation workflows/actions
- moves helm charts into `chart/` directory
- update helm chart to be inline with other services
- updated skaffold to use helm chart
- Remove old kubernetes manifests


chart/templates/deployment.yaml
```diff
@@ -108,14 +106,14 @@ spec:
...
+              cpu: "25m"
             limits:
-              memory: "128Mi"
-              cpu: "500m"
+              memory: "32Mi"
+              cpu: "100m"
           env:
             - name: PUBSUB_TOPIC
-              value: 'sh.keptn.event.test.triggered'
+              value: 'sh.keptn.>'
```
yeahservice, line 116: Should the PUBSUB_TOPIC also be configureable in `.Values`, since `sh.keptn.>` won't work when the job-executor runs in an execution plane?

yeahservice (dismissed):
Awesome stuff 🚀

chart/templates/deployment.yaml
christian-kreuzberger-dtx: Thanks for bringing this up, I wasn't aware of this limitation of the execution plane.

.github/actions/unit-tests/action.yml
```diff
@@ -0,0 +1,18 @@
...
+runs:
+  using: "composite"
+  steps: 
+    - name: Install Go
+      uses: actions/setup-go@v2.1.4
+      with:
+        go-version: ${{ inputs.GO_VERSION }}
+    - name: Test
+      shell: bash
+      run: go test -coverprofile=coverage.txt -covermode=atomic -v ./...
```
mowies, line 18: It doesn't seem that you're making use of the coverage file after generating it. I would suggest to either make use of it in some way or to remove the coverage generation.

.github/workflows/CI.yml
```diff
@@ -97,93 +97,121 @@ jobs:
...
+            version=${{ env.VERSION }}
+          REGISTRY_USER: ${{ secrets.REGISTRY_USER }}
+          REGISTRY_PASSWORD: ${{ secrets.REGISTRY_PASSWORD }}
+          DOCKERFILE: ${{ env.DOCKERFILE_INITCONTAINER }}
+          PUSH: ${{(github.event_name == 'push' || github.event_name == 'workflow_dispatch' || github.event.pull_request.head.repo.full_name == github.repository)}}
+
+      - id: report_docker_build_to_pr
+        name: Report Docker Build to PR
+        if: (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository)
+        uses: marocchino/sticky-pull-request-comment@v2.2.0
```
mowies, line 160: This could be updated to use `v2` instead

mowies (changes requested):

chart/templates/deployment.yaml
christian-kreuzberger-dtx: I have clarified with @johannes-b @agrimmer and @thisthat that this is a known limitation of Keptn execution-plane.

Therefore I will, as you suggested, create a config entry for it, so it can be overwritten using helm values.

.github/workflows/CI.yml
christian-kreuzberger-dtx: fixed

.github/actions/unit-tests/action.yml
christian-kreuzberger-dtx: fixed

chart/templates/deployment.yaml
christian-kreuzberger-dtx: done

mowies (approved):

Issue #127, opened 2021-12-17
Add support for imagePullPolicy


christian-kreuzberger-dtx:
Please add support for `imagePullPolicy` for Kubernetes Jobs - see https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy for details.

**Example**:

```
apiVersion: v2
actions:
  - name: "Run locust"
    events: ...
    tasks:
      - name: "Run locust tests"
        files:
          - locust/basic.py
          - locust/import.py
          - locust/locust.conf
        image: "locustio/locust"
        imagePullPolicy: "Always"
        cmd:
          - locust
```

Please adhere to the [Default IMage pull policy](https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting) by Kubernetes.

# Definition of Done

- [ ] Image Pull Policy field has been added in code
- [ ] README.md contains infos about this field

PR #135, opened 2021-12-22
feat(core): support imagePullPolicy for tasks(#127)

Paolo-Chila:
Introduce support for specifying imagePullPolicy for tasks in
configuration.
No validation of the value is performed on client side,
  we are relying on kubernetes apiserver to perform all the necessary
  validations and apply sensible default as specified in
  https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy


christian-kreuzberger-dtx (approved):
LGTM

PR #139, opened 2021-12-29
chore: Enable skaffold to build and deploy initcontainer

christian-kreuzberger-dtx:
## This PR

- enables skaffold to also build and deploy the initcontainer
- added a `rollme` annotation to the deployment of job-executor-service such that it always restarts, even when just helm values are updated (and not the Kubernetes image itself)
- Update `Dockerfile` and `initcontainer.Dockerfile` to look the same

## Proof

```
skaffold run --tail
```

image

```
keptn trigger delivery ...
```

image

image

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,7 +3,10 @@ README.md
 .idea
 skaffold.yaml
 reviewdog.yml
-charts/
+chart/
 deploy/
 .github/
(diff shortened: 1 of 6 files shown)


Paolo-Chila (approved):
LGTM

Issue #154, opened 2022-01-21
pods of jobs are not automatically deleted


christian-kreuzberger-dtx:
Using job-executor-service 0.1.5, jobs are cleaned up/deleted, but pods are not, e.g.:

```bash
$ kubectl -n keptn get pods | grep job
job-executor-service-job-045986d5-5342-4673-9df9-4311-1-znxw2     0/1     Completed    0          6d18h
job-executor-service-job-04ac7243-0dfb-4320-98ba-57b7-1-gz59b     0/1     Error        0          21d
job-executor-service-job-0852acb3-c40e-4698-90d8-11ef-1-5gwjk     0/1     Completed    0          22d
job-executor-service-job-087f737b-d86a-417c-b16f-e064-1-2b862     0/1     Init:Error   0          22d
job-executor-service-job-08efd292-9734-4e4c-91fe-01ab-1-tvrtx     0/1     Init:Error   0          30d
job-executor-service-job-092f1737-01ff-4847-a7e1-65e8-1-mb7qf     0/1     StartError   0          22d
job-executor-service-job-0f37fe2a-13ff-4e77-b7f8-cfc6-1-t775s     0/1     Completed    0          22d
job-executor-service-job-10af78b3-b3e0-4744-a7e5-d116-1-znzn4     0/1     Completed    0          22d
job-executor-service-job-114adf28-5eba-42c9-a5c3-76c1-1-nn79b     0/1     Completed    0          21d
job-executor-service-job-12efe30b-8cc4-47c8-a40e-7f54-1-fj74l     0/1     Init:Error   0          22d
job-executor-service-job-138d54b5-1fd7-4c4d-aafd-9310-1-jn5xn     0/1     Init:Error   0          21d
job-executor-service-job-13af637e-4846-43a4-932b-eb4a-1-hmjmb     0/1     Completed    0          21d
job-executor-service-job-14752fc9-4709-4daf-9eda-7fb1-1-mhw7g     0/1     Error        0          22d
job-executor-service-job-151c78ff-c80b-4505-9bf6-c2ba-1-kdks2     0/1     Completed    0          21d
job-executor-service-job-1d1e5c86-146b-4b9b-bccd-c86a-1-l8ztr     0/1     Completed    0          22d
job-executor-service-job-1de90dcb-f1f5-4af2-998e-b4e1-1-cb827     0/1     Init:Error   0          22d
...
```

[...]

PR #155, opened 2022-01-24
fix: clean-up pods when deleting kubernetes jobs (#154)

pchila:


christian-kreuzberger-dtx (approved):
Tried it out, works as intended now :) thanks!

LGTM

PR #162, opened 2022-01-27
(fix): Fixes #161

### LINKED ISSUE #161: Clarify File Locations
Background Slack thread: https://keptn.slack.com/archives/CGS5TA283/p1643267477512509

File locations should be specified relative to the `<service>` root, not (as the readme currently suggests) relative from the `job` folder.

For example, given a file and folder structure of:

```
<service>/job/config.yaml
<service>/job/files/file1.json
```

The correct `files` block syntax is:

```
files:
  - job/files/file1.json
```

Dynatrace-Adam-Gardner:
It's logical that people will store job-executor-service files inside the `job` folder. Therefore file paths should be relative to that folder and not the service root.
Make the docs more clear on this.

Signed-off-by: Adam Gardner <adam.gardner@dynatrace.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,9 +171,9 @@ actions:
     tasks:
       - name: "Run locust tests"
         files:
-          - locust/basic.py
-          - locust/import.py
-          - locust/locust.conf
+          - job/locust/basic.py
+          - job/locust/import.py
+          - job/locust/locust.conf
         image: "locustio/locust"
         cmd:
           - locust
@@ -276,17 +276,17 @@ The configuration contains the following section:
 tasks:
   - name: "Run locust tests"
     files:
-      - locust/basic.py
-      - locust/import.py
-      - locust/locust.conf
+      - job/locust/basic.py
+      - job/locust/import.py
+      - job/locust/locust.conf
     image: "locustio/locust"
     cmd:
       - locust
     args:
       - '--config'
-      - /keptn/locust/locust.conf
+      - /keptn/job/locust/locust.conf
       - '-f'
-      - /keptn/locust/basic.py
+      - /keptn/job/locust/basic.py
       - '--host'
       - $(HOST)
 ```
@@ -314,9 +314,9 @@ cmd:
   - locust
 args:
   - '--config'
-  - /keptn/locust/locust.conf
+  - /keptn/job/locust/locust.conf
   - '-f'
-  - /keptn/locust/basic.py
+  - /keptn/job/locust/basic.py
   - '--host'
   - $(HOST)
 env:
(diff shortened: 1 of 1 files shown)


christian-kreuzberger-dtx:
Hi Adam, I don't understand what you mean here.
Files are stored in the `/keptn` folder, not in the `/keptn/job` folder.

Dynatrace-Adam-Gardner:
Most Keptn services create a folder structure like: `<KeptnService>/<ServiceFolder>/files` like `service1/dynatrace/sli.yaml`

Therefore I think most users would assume the job executor service follows the same pattern and thus `service1/job/myFile.yaml`

When this file is copied into the container, the `job` folder is copied into the `keptn` folder so inside the container, the path becomes: `/keptn/job/myFile.yaml`

christian-kreuzberger-dtx:
Okay, I think the general problem is that it's not documented how you should add the respective files. 

We only documented how to add the job config, which goes to `<service>/job/config.yaml`, e.g.:
```bash
keptn add-resource --project=$PROJECT --service=helloservice --stage=qa --resource=job-config.yaml --resourceUri=job/config.yaml
```

But we didn't specify where to put `locust/basic.py` etc..., which should also be `<service>/locust/basic.py`, e.g.:

```bash
keptn add-resource --project=$PROJECT --service=helloservice --stage=qa --resource=./locust/basic.py
keptn add-resource --project=$PROJECT --service=helloservice --stage=qa --resource=./locust/locust.conf
```

and therefore
```yaml
        files:
          - locust/basic.py
          - locust/import.py
          - locust/locust.conf
```
is correct.


You shouldn't add them into the `job` directory, but obviously, there is nothing preventing you from doing that.
For the sake of simplicity, however, I would advise to keep it the way it's in README.

What's your oppinion on that @Dynatrace-Adam-Gardner ?

Also pinging @thschue and @yeahservice  on their opinion :)

agardnerIT:
`You shouldn't add them into the job directory` Why? It makes perfect sense to me that all job-executor-service files are stored in the `job` folder, just like all `dynatrace-service` files are stored under the `dynatrace` folder.

So I'd like all readme's to assume that teh base directory is actually `job`.

thschue:
I agree with @christian-kreuzberger-dtx that it's more straightforward to keep it as it is. The job executor is more or less a unique integration, which can act as a drop-in replacement for many other integrations (helm-service, monaco-service, argo-service). In some setups, it is the only execution plane service used. Keeping the top-level on the service folder allows replacing some services in place and holding the files inside the job folder. I prefer not to keep the job-specific folders (e.g., helm charts) in the job folder.

christian-kreuzberger-dtx:
Please think about it like this: You don't store all Keptn Core related stuff in a "keptn" sub folder either.
"job-executor" is a tool that helps you to integrate with other tools, e.g., helm, jmeter, locust, kubectl, ...

We are not restricting the user, they can always put stuff in the "job" folder, but it's not the default and not our recommended best practice.
Closing this PR for now.

Issue #159, opened 2022-01-27
API Token Regex Incorrect


Dynatrace-Adam-Gardner:
Keptn API Tokens can contain `+` so deploying via helm fails when API tokens contain forbidden chars.

https://github.com/keptn-contrib/job-executor-service/blob/b3eeccef1da6f8dfa873916188ed470a6463b56f/chart/values.schema.json#L44

pchila:
Thanks @Dynatrace-Adam-Gardner for the spotting the issue and providing the fix :) PR merged

PR #165, opened 2022-01-27
fix: Revert "Allow '+' in API Token helm value (#160) (#159)"

pchila:
This reverts commit ed463da98fb50b90fa485483f81505bea49855dd.

Still not clear how a '+' character got into the keptn api token.
Needs more investigation about the root cause rather than a quick fix.
Reverting the change till we know more.
Reverts #160

### DIFF
--- a/chart/values.schema.json
+++ b/chart/values.schema.json
@@ -41,7 +41,7 @@
                 "type": "boolean"
               },
               "token": {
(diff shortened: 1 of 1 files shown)


christian-kreuzberger-dtx (approved):
LGTM

Issue #95, opened 2021-11-10
Silent mode for events in non-remote execution planes does not work


yeahservice:
The following if prevents processing of any non `.triggered` events and therefore the silent mode can't really be used.

```
	if !strings.Contains(event.Type(), ".triggered") {
		return nil
	}
```
see https://github.com/keptn-contrib/job-executor-service/blob/be2956e0aeebeaa5c69b883131ca2dee91b235ba/cmd/job-executor-service/main.go#L86

It would probably be best to remove the if and completely test the silent mode.

TannerGilbert:
I'll work on this

PR #182, opened 2022-03-01
fix: Allow handling of event types other than *.triggered

pchila:
Removed code that skipped any event that did not contain '.triggered' in
the type. This will allow the usage of 'silent mode' when matching event
types different from '*.triggered'


TannerGilbert (approved):
LGTM. Thanks @pchila this fixes #95.

christian-kreuzberger-dtx (approved):
LGTM

PR #185, opened 2022-03-02
feat: Add labels to environment variables

christian-kreuzberger-dtx:
Signed-off-by: Christian Kreuzberger <christian.kreuzberger@dynatrace.com>


thschue (approved):
lgtm

pkg/k8sutils/job.go
```diff
@@ -282,6 +283,21 @@ func (k8s *k8sImpl) prepareJobEnv(task config.Task, eventData *keptnv2.EventData
...
+		// replace - with _
+		key = strings.Replace(key, "-", "_", -1)
+		// uppercase string
+		key = strings.ToUpper(key)
+
+		jobEnv = append(jobEnv,
+			v1.EnvVar{
+				Name:  "LABELS_" + key,
+				Value: value,
+			},
```
pchila, line 297: Some small optimization using strings.Builder and strings.Replacer
```suggestion
      	sb := new(strings.Builder)
      	sr := strings.NewReplacer("-", "_", " ", "_")
	// append labels as environment variables
	for key, value := range eventData.Labels {
	       sb.Reset()
	       sb.WriteString("LABEL_")
	       sr.WriteString(sb, strings.ToUpper(key))
		
		jobEnv = append(jobEnv,
			v1.EnvVar{
				Name:  sb.String(),
				Value: value,
			},
```

pkg/k8sutils/job_test.go
```diff
@@ -111,23 +112,18 @@ func TestPrepareJobEnvFromEvent(t *testing.T) {
...
-
-	assert.Equal(t, jobEnv[0].Name, "HOST")
-	assert.Equal(t, jobEnv[0].Value, "https://keptn.sh")
-
-	assert.Equal(t, jobEnv[1].Name, "DEPLOYMENT_STRATEGY")
-	assert.Equal(t, jobEnv[1].Value, "user_managed")
-
-	assert.Equal(t, jobEnv[2].Name, "TEST_STRATEGY")
-	assert.Equal(t, jobEnv[2].Value, "health")
+	assert.NoErrorf(t, err, "prepareJobEnv should not return an error")
```
pchila, line 126:
```suggestion
	require.NoError(t, err)
```

pchila (dismissed):
A couple of minor changes to improve clarity and maybe allocate fewer strings ;)
Great to see nice and clear assertions using stretcher/testify

pkg/k8sutils/job.go
christian-kreuzberger-dtx: The replacer I can see how it helps with performance, especially when we have a lot of labels. Also thanks for adding the replacement of space with `_`.

However, I'm not seeing a benefit of using string builder - code looks more complicated than it should be. Given that we are only concatenating  `LABEL_` with the output of string replacer, I would keep it like this:
```
key = "LABELS_" + sr.Replace(sb, strings.ToUpper(key))
```

btw, did you change `LABELS` to `LABEL` on purpose?

pchila: the change from `LABELS` to `LABEL` is due to reading quickly :sweat_smile: 
In general strings.Builder is indicated when concatenating/manipulating strings but in this case it's probably equivalent due to `strings.ToUpper` working only on strings (so those allocation are difficult to get rid of). None of my suggested changes impact heavily the performance, it's more a good reminder that strings.builder exists and can be used ;)

christian-kreuzberger-dtx: did my changes

pchila (approved):
LGTM

PR #187, opened 2022-03-03
build: Upgrade to go 1.17

christian-kreuzberger-dtx:
* Fix an issue with queueGroup in helm chart
* Upgrade to Go 1.17

*Note* from https://go.dev/doc/go1.17:
> If a module specifies go 1.17 or higher, the module graph includes only the immediate dependencies of other go 1.17 modules, not their full transitive dependencies. (See [Module graph pruning](https://go.dev/ref/mod#graph-pruning) for more detail.)
>  For the go command to correctly resolve transitive imports using the pruned module graph, the go.mod file for each module needs to include more detail about the transitive dependencies relevant to that module. If a module specifies go 1.17 or higher in its go.mod file, its go.mod file now contains an explicit require directive for every module that provides a transitively-imported package. (In previous versions, the go.mod file typically only included explicit requirements for directly-imported packages.)


pchila (dismissed):
LGTM

pchila (approved):
LGTM

PR #197, opened 2022-03-21
feat: Implement liveness and readiness endpoints

### LINKED ISSUE #174: Add liveness/readiness probes for job-executor service
In PR #173 The old liveness and readiness probes had to be removed since the distributor is only listening on `localhost` and job-executor doesn't spin up an HTTP server so the `/health` endpoint handled by the distributor is not reachable (the pod would be continuously restarted by k8s due to the probe failure).

We need to introduce an endpoint reachable from k8s that is able to check for both `job-executor` and `distributor` (maybe this one can be done through a forward to `http://localhost:8081/health`) liveness and readiness.

/cc @agrimmer @christian-kreuzberger-dtx

TannerGabriel:
Implement liveness and readiness endpoints and use them in Helm files.

## Fixes
#174

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -34,6 +34,14 @@ spec:
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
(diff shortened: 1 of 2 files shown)


christian-kreuzberger-dtx (approved):
LGTM

PR #204, opened 2022-03-25
feat: Restrict service account of jobs

Raffy23:
## This PR
* Introduces the ability to change the service account for jobs started by the job-executor-service with a more restrictive default service account


chart/templates/jobserviceaccount.yaml
```diff
@@ -0,0 +1,35 @@
...
+  {{- toYaml . | nindent 4 }}
+  {{- end }}
+---
+# Role for the jobs
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: job-executor-service_job
+  namespace: {{ .Release.Namespace }}
+rules: []
```
pchila, line 19: I am wondering if we want to give the option to customize the set of permissions when performing `helm install/upgrade` similar to what we want to do with the security context :thinking: 
It would basically mean having the array of rules in values.yaml (empty by default but users can customize it when installing the JES)
Here the feedback of @christian-kreuzberger-dtx  would be very appreciated :smile:

cmd/job-executor-service/main.go
```diff
@@ -160,6 +163,11 @@ func _main(args []string, env envConfig) int {
 		keptnOptions.UseLocalFileSystem = true
 	}
 
+	// Checking if the given job service account is empty
+	if env.DefaultJobServiceAccount == "" {
+		log.Println("WARNING: Using default service account for jobs!")
```
pchila, line 168: We can try to make it clearer for the user since there are multiple defaults in play (at least they should have a hint at what they need to change to avoid it)
```suggestion
		log.Println("WARNING: No default service account for jobs configured: using kubernetes default service account!")
```

pkg/k8sutils/job.go
```diff
@@ -75,7 +76,7 @@ func (k8s *k8sImpl) CreateK8sJob(
 	automountServiceAccountToken := jobSettings.EnableKubernetesAPIAccess
 
 	// specify empty service account name for job
-	serviceAccountName := ""
+	serviceAccountName := jobSettings.DefaultJobServiceAccount
```
pchila, line 79: Couple of things here:
- if we keep line 79 we should probably delete the comment in line 78 :wink: 
- Aren't we supposed to be backward-compatible?  If  we have `jobConfig.enableKubernetesApiAccess=true` we should specify the default service account

pchila (changes requested):
We are missing a bit of backwards compatibility here. Would you mind adding that part ?

pkg/k8sutils/job.go
Raffy23: This change doesn't affect the code path that will be taken if `jobConfig.enableKubernetesApiAccess=true`:
https://github.com/keptn-contrib/job-executor-service/blob/feat/restrict-job-service-account/pkg/k8sutils/job.go#L77-L83.

The service account will be overwritten as it is the case currently in the master branch. But a helm upgrade with `--reuse-values` might break backwards compatibility, because the new service account will be used.

chart/templates/jobserviceaccount.yaml
Raffy23: If we let the user configure the service accounts, I would prefer to have two different service accounts. One with an empty set of rules for all jobs that don't need access to the Kubernetes APIs and another one that can be configured when installing JES. The additional service account can be used for jobs if `jobConfig.enableKubernetesApiAccess=true` is used instead of using the JES service account.

christian-kreuzberger-dtx: For now the empty rules array is perfectly fine as the default setting. The goal is to restrict default access of workloads. 
There is another story where we will allow users to overwrite the service account for specific workloads.

christian-kreuzberger-dtx (approved):
LGTM from my side. @pchila I believe we can merge this for now

pchila (approved):

PR #207, opened 2022-03-30
feat!: Move job-executor-service to it's own namespace

Raffy23:
## This PR
* moves the job-executor-service into it's own namespace (e.g.: `keptn-jes`) to isolate the jobs from other Keptn services
* moves the Keptn API token into a secret and the Keptn API endpoint to a configmap
* restricts the service account to only include required permissions


README.md
```diff
@@ -37,46 +37,46 @@ Please note: Newer Keptn versions might be compatible, but compatibility has not
...
-You can install job-executor-service either in the same cluster and namespace as Keptn (see above), or on a completely 
+The *job-executor-service* can be installed as a part of [Keptn's uniform](https://keptn.sh) using `helm`.
+It is recommended to install the *job-executor-service* on the remote execution-plane, which can be on the same cluster, or on a completely
 separate Kubernetes environment (see [Keptn docs: Multi-cluster setup](https://keptn.sh/docs/0.11.x/operate/multi_cluster/) for details).
+Alternatively it is also possible to install the *job-executor-service* in the [same namespace as Keptn](#installation-in-the-same-namespace-as-keptn). 
 
-In order to do the latter, job-executor-service helm chart provides some values that can be configured:
-
-* `remoteControlPlane.enabled` - needs to be set to `true` in order to connect to the remote Keptn instance
+In order to install the *job-executor-service* on the remote execution plane, some values of the helm chart can be configured:
```
christian-kreuzberger-dtx, line 45: Replace "some values of the helm chart can be configured" with "some values of the helm chart need to be configured"

README.md
```diff
@@ -37,46 +37,46 @@ Please note: Newer Keptn versions might be compatible, but compatibility has not
...
 * `remoteControlPlane.topicSubscription` - list of Keptn CloudEvent types that this instance should listen to, e.g., `sh.keptn.event.remote-task.triggered`
 * `remoteControlPlane.api.protocol` - protocol (`http` or `https`) used to connect to the remote control plane
-* `remoteControlPlane.api.hostname` - Keptn API Hostname (e.g., `1.2.3.4.nip.io`)
+* `remoteControlPlane.api.hostname` - Keptn API Hostname (e.g., `1.2.3.4.nip.io` / `api-gateway-nginx.keptn`)
```
christian-kreuzberger-dtx, line 48: Maybe: "`api-gateway-nginx.keptn` in case you want to access a Keptn installation on the same Kubernetes cluster"

README.md
```diff
@@ -37,46 +37,46 @@ Please note: Newer Keptn versions might be compatible, but compatibility has not
...
 * `remoteControlPlane.api.token` - Keptn API Token (can be obtained from Bridge)
 
 **Example**
 ```bash
 KEPTN_API_PROTOCOL=http # or https
 KEPTN_API_HOST=<INSERT-YOUR-HOSTNAME-HERE> # e.g., 1.2.3.4.nip.io
- KEPTN_API_TOKEN=<INSERT-YOUR-KEPTN-API-TOKEN-HERE>
+KEPTN_API_TOKEN=<INSERT-YOUR-KEPTN-API-TOKEN-HERE>
```
christian-kreuzberger-dtx, line 55: the space is here on purpose, such that the api token is not put into the history of your terminal

README.md
```diff
@@ -37,46 +37,46 @@ Please note: Newer Keptn versions might be compatible, but compatibility has not
...
+### Installation in the same namespace as Keptn
+
+To install the *job-executor-service* in the same namespace as Keptn you have to disable the `remoteControlPlane.enabled` value as shown below.
+But be aware that all jobs will be executed in this namespace.
+
+```bash
+helm upgrade --install --create-namespace -n <NAMESPACE> \
+  job-executor-service https://github.com/keptn-contrib/job-executor-service/releases/download/<VERSION>/job-executor-service-<VERSION>.tgz \
+ --set remoteControlPlane.enabled=false
+```
```
christian-kreuzberger-dtx, line 78: I think we should remove this option. Installing it on the same namespace should not be possible, especially in the context of Kubernetes hardening, separation makes sense.

chart/templates/_helpers.tpl
```diff
@@ -60,3 +60,11 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{- define "job-executor-service.remote-conrol-plane.token" -}}
```
christian-kreuzberger-dtx, line 64: typo: `conrol` -> `control`

chart/templates/configmap.yaml
```diff
@@ -11,3 +11,19 @@ data:
   default_resource_requests_memory: "128Mi"
   always_send_finished_event: "false"
   enable_kubernetes_api_access: "{{ .Values.jobConfig.enableKubernetesApiAccess | default false }}"
+  keptn_api_endpoint: {{ (printf "%s://%s/api" .Values.remoteControlPlane.api.protocol .Values.remoteControlPlane.api.hostname) }}
+  init_container_configuration_endpoint:
+    {{- if .Values.remoteControlPlane.enabled }}
+      {{ (printf "%s://%s/api/configuration-service" .Values.remoteControlPlane.api.protocol .Values.remoteControlPlane.api.hostname) }}
+    {{- else }}
+      "http://configuration-service:8080"
+    {{- end }}
```
christian-kreuzberger-dtx, line 20: Let's also remove the if else here. REmote control plane should be the default, end we no longer access configurations-ervice directly.

chart/templates/configmap.yaml
```diff
@@ -11,3 +11,19 @@ data:
...
+---
+# Secret for the Keptn API Token
+apiVersion: v1
+kind: Secret
+type: Opaque
+metadata:
+  name: job-serivce-keptn-secrets
+data:
+  token: {{ b64enc (include "job-executor-service.remote-conrol-plane.token" .) | quote }}
```
christian-kreuzberger-dtx, line 29: I think we should separate the secret into its own manifest file, what do you think?

chart/templates/deployment.yaml
```diff
@@ -165,9 +161,15 @@ spec:
...
-              value: "{{ .Values.remoteControlPlane.api.protocol }}://{{ .Values.remoteControlPlane.api.hostname }}/api"
+              valueFrom:
+                configMapKeyRef:
+                  name: job-service-config
+                  key: keptn_api_endpoint
             - name: KEPTN_API_TOKEN
-              value: "{{ .Values.remoteControlPlane.api.token }}"
+              valueFrom:
+                secretKeyRef:
+                  name: job-serivce-keptn-secrets
```
christian-kreuzberger-dtx, line 171: typo: `serivce` -> `service`

chart/templates/serviceaccount.yaml
```diff
@@ -17,30 +17,15 @@ metadata:
...
-      - "*"
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["list"]
+  - apiGroups: [""]
+    resources: ["secrets", "pods/log"]
+    verbs: ["get"]
+  - apiGroups: ["batch"]
+    resources: ["jobs"]
+    verbs: ["get", "create", "watch"]
```
christian-kreuzberger-dtx, line 28: do we need this change? I'd rather have the longer list as it's easier to read

chart/values.yaml
```diff
@@ -24,11 +24,11 @@ distributor:
       enabled: true                          # Enable connection via Nats queue group to support exactly-once message processing
 
 remoteControlPlane:
-  enabled: false                             # Enables remote execution plane mode
+  enabled: true                              # Enables remote execution plane mode
```
christian-kreuzberger-dtx, line 27: I think we should remove the `enabled` flag. It's always enabled, and we require connection to the Rest API + token

pkg/k8sutils/job.go
```diff
@@ -127,12 +126,26 @@ func (k8s *k8sImpl) CreateK8sJob(
...
-									Value: jobSettings.InitContainerConfigurationServiceAPIEndpoint,
+									Name: "CONFIGURATION_SERVICE",
+									ValueFrom: &v1.EnvVarSource{
+										ConfigMapKeyRef: &v1.ConfigMapKeySelector{
+											LocalObjectReference: v1.LocalObjectReference{
+												Name: "job-service-config",
+											},
+											Key: "init_container_configuration_endpoint",
+										},
+									},
```
christian-kreuzberger-dtx, line 137: This means that the underlying job needs to be able to access the configmap - I'm not sure if we want that

pkg/k8sutils/job.go
```diff
@@ -127,12 +126,26 @@ func (k8s *k8sImpl) CreateK8sJob(
...
-									Value: jobSettings.KeptnAPIToken,
+									Name: "KEPTN_API_TOKEN",
+									ValueFrom: &v1.EnvVarSource{
+										SecretKeyRef: &v1.SecretKeySelector{
+											LocalObjectReference: v1.LocalObjectReference{
+												Name: "job-serivce-keptn-secrets",
+											},
+											Key: "token",
+										},
+									},
```
christian-kreuzberger-dtx, line 148: Same problem with the secret. I am in favour of not having the secret directly in the manifest though.

christian-kreuzberger-dtx (changes requested):
Left a lot of comments :D 

Some general remarks:
* I think with this hardening we should make remoteControlPlane the only available setting (this is a breaking change, I know).
* We should allow installation on the same cluster, but in a different namespace (we shouldn't restrict the namespace name though, just provide the best practice by default). This means we can have `api-gateway-nginx.keptn` as the default API endpoint, but we still require the user to provide the token
* I am not sure about the implications of not having certain environment variables directly set for the jobs - though I think it means that the job needs access to the configmap and secret, and I'm not sure if that's the best approach (but I'm willing to take this in order to not have the values directly in the manifest).

pkg/k8sutils/job.go
Raffy23: I tested it with jobs that had the following service account (https://github.com/keptn-contrib/job-executor-service/blob/4e9505a03c5a53c3349a9ce5e75a432a0f2b2265/chart/templates/jobserviceaccount.yaml) and it worked. So I'm not exactly sure how the permissions are for these references.

chart/templates/configmap.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: job-service-config
+  namespace: {{ .Release.Namespace }}
```
christian-kreuzberger-dtx, line 5: Is this required? AFAIK you don't need to specify the namespace here, it's implied by `helm install -n <NAMESPACE>`.
@pchila please correct me if I'm wrong :D

chart/templates/deployment.yaml
```diff
@@ -125,14 +121,11 @@ spec:
               cpu: "100m"
           env:
             # Note: in the case of remote execution plane we need to specify a list of CloudEvent types
-            {{- if .Values.remoteControlPlane.enabled }}
             - name: PUBSUB_TOPIC
               value: "{{ .Values.remoteControlPlane.topicSubscription }}"
-            {{- else }}
             # Note: In case of "normal" setup (within the same Kuberntes cluster as Keptn control-plane) it is enough to subscribe to sh.keptn.>
             - name: PUBSUB_TOPIC
               value: "sh.keptn.>" # all CloudEvent types
```
christian-kreuzberger-dtx, line 128: I think we can remove the "generic" PUBSUB_TOPIC now. Also "sh.keptn.>" won't work in remote execution plane setups.

chart/templates/secrets.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+type: Opaque
+metadata:
+  name: job-service-keptn-secrets
+  namespace: {{ .Release.Namespace }}
```
christian-kreuzberger-dtx, line 6: Same here, not sure if we actually need to specify the namespace

pkg/k8sutils/job.go
christian-kreuzberger-dtx: Confirmed, this works but doesn't provide the underlying job access to the kubernetes api, so everything works nicely

chart/templates/configmap.yaml
Raffy23: No it's not required, helm will generate the appropriate metadata for the target namespace. Not sure why I put it there ...

pkg/k8sutils/job.go
Raffy23: No additional permissions are necessary for the job pod, because we just reference the secret and aren't asking over the API endpoint

christian-kreuzberger-dtx (dismissed):
LGTM
@pchila can you also take a look

chart/templates/configmap.yaml
```diff
@@ -11,4 +11,6 @@ data:
   default_resource_requests_memory: "128Mi"
   always_send_finished_event: "false"
   enable_kubernetes_api_access: "{{ .Values.jobConfig.enableKubernetesApiAccess | default false }}"
+  keptn_api_endpoint: {{ (printf "%s://%s/api" .Values.remoteControlPlane.api.protocol .Values.remoteControlPlane.api.hostname) }}
```
pchila, line 14: are we sure we are always going to use default protocol ports ? Or the port is to be included in the hostname ?

gh-actions-scripts/build_helm_chart.sh
```diff
@@ -42,7 +42,7 @@ fi
 mv "${IMAGE}-${VERSION}.tgz" "installer/${IMAGE}-${VERSION}.tgz"
 
 #verify the chart
-helm template "installer/${IMAGE}-${VERSION}.tgz" --dry-run > /dev/null
+helm template "installer/${IMAGE}-${VERSION}.tgz" --dry-run --set remoteControlPlane.api.token=thisisnotavalidkeptntoken > /dev/null
```
pchila, line 45: If we want to verify the chart why didn't we use a [helm lint](https://helm.sh/docs/helm/helm_lint/) command before packaging ?  I know it didn't change in this PR, just curious :thinking:

Raffy23: Didn't know that linting existed, I will adapt the current script to use it ... the edit was a pretty ugly fix to get things working

chart/templates/configmap.yaml
christian-kreuzberger-dtx: Port is part of hostname

gh-actions-scripts/build_helm_chart.sh
christian-kreuzberger-dtx: The initial thought here (coming from Keptn core, waaaay back...) was to verify that the helm chart can be rendered. Not sure if `helm lint` also does the job.

chart/values.yaml
```diff
@@ -27,6 +27,7 @@ remoteControlPlane:
   topicSubscription: "sh.keptn.event.test.triggered,sh.keptn.event.deployment.triggered,sh.keptn.event.evaluation.triggered,sh.keptn.event.release.triggered,sh.keptn.event.approval.triggered" # Specify list of CloudEvents to subscribe to in the case of remote control plane
   api:
     protocol: "http"                         # Used Protocol (http, https)
+    # port: 80                               # (Optional) If the API is reachable over a non-standard port
     hostname: "api-gateway-nginx.keptn"      # Hostname of the control plane cluster (and Port)
```
christian-kreuzberger-dtx, line 31: In the case of adding port as a value, the documentation of `hostname` does not work (says "and Port")

Raffy23: I reverted that commit, I think it's better to stay with the old solution by just adding the Port to the `hostname`

gh-actions-scripts/build_helm_chart.sh
pchila: `helm lint` checks that the chart is formally correct, `helm template` shows you what the content of the rendered chart would be but we are throwing away those to `/dev/null`
I am not saying that we need to change, I was just wondering why one was used instead of the other...

pchila (approved):

christian-kreuzberger-dtx (approved):
LGTM

Issue #209, opened 2022-03-31
Allow auto-detection of Keptn API endpoint and API token during helm install in same Kubernetes cluster


christian-kreuzberger-dtx:
With PR https://github.com/keptn-contrib/job-executor-service/pull/207 we have removed the ability to install job-executor-service "directly" next to a Keptn control-plane in the same namespace (which allows keptn/distributor to directly communicate with NATS).

As an easy workaround, a user can install job-executor-service in the same namespace, but communicate via `http://api-gateway-nginx` (or `http://api-gateway-nginx.<NAMESPACE>` and the api-token from `kubectl get secret keptn-api-token -n keptn -ojsonpath='{.data.keptn-api-token}' | base64 -d`), e.g.:

```
helm install ... --set remoteControlPlane.api.hostname=api-gateway-nginx.keptn,remoteControlPlane.api.token=$(kubectl get secret keptn-api-token -n keptn -ojsonpath='{.data.keptn-api-token}' | base64 -d)
```
(Note that this does only work when you are using `bash` or similar)

However, **as a user**, I would like to have an autodetect feature for easy installation on the same Kubernetes cluster, e.g.,

```
helm install ... --set remoteControlPlane.autoDetect.enabled=true,remoteControlPlane.autoDetect.namespace=keptn
```

which would automatically configure `remoteControlPlane.api.token`, `remoteControlPlane.api.protocol` and `remoteControlPlane.api.hostname`.

**values.yaml example**
```yaml
remoteControlPlane:
  topicSubscription: "sh.keptn.event.test.triggered,sh.keptn.event.deployment.triggered,sh.keptn.event.evaluation.triggered,sh.keptn.event.release.triggered,sh.keptn.event.approval.triggered" # Specify list of CloudEvents to subscribe to in the case of remote control plane
  api:
    protocol: ""                             # Used Protocol (http, https)
    hostname: ""                             # Hostname of the control plane cluster (and Port)
    apiValidateTls: true                     # Defines if the control plane certificate should be validated
    token: ""                                # Keptn API Token
  autoDetect:                              # auto-detect protocol, hostname and token if control-plane is in same cluster
    enabled: true                          # if set to true, helm tries to auto-detect an existing Keptn install
    namespace: "keptn"                     # in the given namespace, and automatically sets hostname and token
```

**Note**: 
* If `api.token` is set, it must be used (**fail with an error**)
* If `api.protocol` is set, it must be used (**fail with an error**)
[...]

PR #212, opened 2022-04-01
build: Add api-token to skaffold.yaml for better dev experience

christian-kreuzberger-dtx:
Until https://github.com/keptn-contrib/job-executor-service/issues/209 gets implemented, we can leverage this simple change to have an API token set as an environment variable when running skaffold.

See https://skaffold.dev/docs/environment/templating/ for more infos on templating with skaffold.

Example to run:

```bash
KEPTN_API_TOKEN=abcd skaffold run --tail
```

Or even better
```bash
KEPTN_API_TOKEN=$(kubectl get secret keptn-api-token -n keptn -ojsonpath='{.data.keptn-api-token}' | base64 -d) skaffold run --tail
```

### DIFF
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -29,6 +29,10 @@ deploy:
             image: keptncontrib/job-executor-service-initcontainer
         imageStrategy:
           helm: { }
+        setValueTemplates:
+          remoteControlPlane:
+            api:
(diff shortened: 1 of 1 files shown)


Raffy23 (approved):
LGTM

PR #213, opened 2022-04-04
feat: Add allowlist for job images

Raffy23:
## This PR
* adds an allow-list to the job-executor-service to limit the images that can be used in job workloads
* the allow-list is separated by `,` and allows wildcards like `*`


cmd/job-executor-service/main.go
```diff
@@ -239,3 +253,22 @@ func endpointNotFoundHandler(w http.ResponseWriter, r *http.Request) {
 		log.Println(err)
 	}
 }
+
+// buildImageAllowList creates a ImageFilterList from a comma separated string that is present as environment variable
+func buildImageAllowList(envVariable string) (*utils.ImageFilterList, error) {
```
christian-kreuzberger-dtx, line 258: would it make sense to a) put this into utils, and b) make it public?

cmd/job-executor-service/main.go
```diff
@@ -186,8 +191,17 @@ func _main(args []string, env envConfig) int {
...
+		log.Fatalf("failed to generate the allowlist, %v", err)
+	}
+
+	processCloudEventFunc := func(ctx context.Context, event cloudevents.Event) error {
+		return processKeptnCloudEvent(ctx, event, imageFilterList)
+	}
+
 	log.Printf("Starting receiver")
-	log.Fatal(c.StartReceiver(ctx, processKeptnCloudEvent))
+	log.Fatal(c.StartReceiver(ctx, processCloudEventFunc))
```
christian-kreuzberger-dtx, line 204: Why renaming the func?

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.17
 require (
 	github.com/PaesslerAG/jsonpath v0.1.1
 	github.com/cloudevents/sdk-go/v2 v2.5.0
+	github.com/gobwas/glob v0.2.3
```
christian-kreuzberger-dtx, line 8: This package is super outdated, I wonder if there is a newer "better" package we can use.
@pchila any recommendations?

pkg/eventhandler/eventhandlers.go
```diff
@@ -135,6 +138,21 @@ func (eh *EventHandler) startK8sJob(k8s k8sutils.K8s, action *config.Action, jso
 		start: time.Now(),
 	}
 
+	// To execute all tasks atomically, we check all images
+	// before we start executing a single task of a job
+	for _, task := range action.Tasks {
+		if !eh.AllowedImages.Contains(task.Image) {
+			log.Printf("Image %s is not in the allowed!\n", task.Image)
```
christian-kreuzberger-dtx, line 145: I would change this message to `"Forbidden: Image %s is not part of allowedImageList.\n"`

pkg/eventhandler/eventhandlers.go
```diff
@@ -135,6 +138,21 @@ func (eh *EventHandler) startK8sJob(k8s k8sutils.K8s, action *config.Action, jso
...
+
+			if !action.Silent {
+				errorText := fmt.Sprintf("Image %s is not allowed", task.Image)
```
christian-kreuzberger-dtx, line 148: Please don't duplicate generating the error message/string

cmd/job-executor-service/main.go
Raffy23: The function must be wrapped, because I added another parameter (the allow list) to it: 
https://github.com/keptn-contrib/job-executor-service/pull/213/files/7e55ed3bca3b2746a21f55a50f9667c5657a80f0#diff-2c166bc577041c1b2a355e339b9fe7e6d7e088b3e3acaaa32957118a29ac1652R199-R201

pkg/utils/imagefilterlist.go
```diff
@@ -0,0 +1,142 @@
...
+	"regexp"
+	"strings"
+)
+
+const defaultContainerRegistry = "docker.io"
+
+// ImageFilterList represents a list of glob filters that can be used to check if an image matches one of these filters
+type ImageFilterList struct {
+	registryMatcher *regexp.Regexp
+	userMatcher     *regexp.Regexp
```
pchila, line 16: I think we can simplify the checks without having specific regex for registry and user (which may or may not be present in the image name for example `alpine`, `ubuntu`, `mongo` etc. ).

I would propose a simpler algorithm like the following:

- check if the image name matches the allow list, if true return without further processing
- if the image does not match, try splitting on the first `/` and check if the first token is a hostname (optionally removing the port) using https://pkg.go.dev/net#LookupHost or https://pkg.go.dev/net#Resolver.LookupHost (depending if we need to specify a context or if we are ok with context.Background)
- if the first token is not a hostname we can try prepending the `docker.io` default for the docker public image registry and try matching against the allow list again (optionally we could try a HEAD http request against the docker registry but it's more expensive and we may incur in docker registry rate limit)

pkg/utils/imagefilterlist.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// transform all patterns to a compiled regex instance
+	globPatterns := make([]glob.Glob, len(patterns))
+	for index, pattern := range patterns {
+
+		// If the pattern is a * we can skip everything and just create an empty list
+		if pattern == "*" {
```
pchila, line 62: Maybe add a warning here since this could be a simple typo that would allow any image (think about an allowlist like `docker.io/*, ghcr.io,*` which doesn't make much sense...)

We could even be stricter and consider such allow list as invalid and only allow stuff like `*` but nothing else...

pkg/utils/imagefilterlist.go
Raffy23: I think checking the hostname and verifying that the image is present at the docker registry is a little bit out of scope, but I look into simplifying the matching of the images

pkg/utils/imagefilterlist_test.go
```diff
@@ -30,52 +30,67 @@ func TestSpecificImagesMustBeAccepted(t *testing.T) {
...
+	assert.False(t, helper.Contains("my-user/my-image:1.2.3"))
+
+	// Do not allow images from other registries
 	assert.False(t, helper.Contains("ghcr.io/my-user/some-image"))
 	assert.False(t, helper.Contains("ghcr.io/my-user/my-image:latest"))
 }
 
-func TestAcceptEverythingFromDefaultRegistry(t *testing.T) {
+func TestWildcardFormatsRegistry(t *testing.T) {
 	allowedImageList := []string{"docker.io/*", "*[AB]/some-image*"}
```
pchila, line 55: Let's add some wildcard on the tag ;)
```suggestion
	allowedImageList := []string{"docker.io/*", "*[AB]/some-image*", "my-docker-mirror.myorg/loadimpact/k6:0.34.*"}
```

pkg/utils/imagefilterlist_test.go
```diff
@@ -30,52 +30,67 @@ func TestSpecificImagesMustBeAccepted(t *testing.T) {
...
 	assert.True(t, helper.Contains("A/some-image:latest"))
-	assert.False(t, helper.Contains("some-image"))
 	assert.True(t, helper.Contains("docker.io/my-user/my-image:1.2.3"))
+	assert.True(t, helper.Contains("ghcr.io/B/some-image"))
 
+	assert.False(t, helper.Contains("some-image"))
+	assert.False(t, helper.Contains("ghcr.io/A/my-image:1.2.3"))
 	assert.False(t, helper.Contains("ghcr.io/my-user/some-image"))
 	assert.False(t, helper.Contains("ghcr.io/my-user/my-image:latest"))
 }
```
pchila, line 69:
```suggestion
        assert.True(t, helper.Contains("my-docker-mirror.myorg/loadimpact/k6:0.34.0"))
        assert.True(t, helper.Contains("my-docker-mirror.myorg/loadimpact/k6:0.34.1"))
        assert.False(t, helper.Contains("my-docker-mirror.myorg/loadimpact/k6:0.35.0"))
        assert.False(t, helper.Contains("my-docker-mirror.myorg/loadimpact/k6"))
}
```

pchila (approved):
LGTM

Issue #124, opened 2021-12-17
Fail with a `.finished` event with proper message and status if initcontainer fails


christian-kreuzberger-dtx:
Right now initcontainer fails when it cannot fetch a file from configuration-service.

For instance, I have this config:
```
apiVersion: v2
actions:
  - name: "Run JMeter"
    events:
      - name: "sh.keptn.event.test.triggered"
        jsonpath:
          property: "$.data.test.teststrategy"
          match: "functional"
      - name: "sh.keptn.event.test.triggered"
        jsonpath:
          property: "$.data.test.teststrategy"
          match: "performance"
    tasks:
      - name: "Run jmeter smoke tests"
        files:
          - jmeter/basiccheck.jmx
          - jmeter/load.jmx
        image: "docker.io/christiankreuzbergerdtx/jmeter:latest"
        ...
```

but unfortunately initcontainer fails with
```
2021/12/17 10:31:18 Error while copying files: could not find file or directory jmeter/basiccheck.jmx for task Run jmeter smoke tests
```

In Keptn Bridge we only get this message:
```
11:31 job-executor-service finished with result fail

Job job-executor-service-job-855e19ed-baed-48b3-973e-0e62-1 failed: job job-executor-service-job-855e19ed-baed-48b3-973e-0e62-1 failed. Reason: BackoffLimitExceeded, Message: Job has reached the specified backoff limit
```

# Definition of Done

- [ ] Keptn Bridge shows the proper error message, e.g., `could not find file or directory jmeter/basiccheck.jmx for task`

Issue #123, opened 2021-12-17
Add troubleshooting step: logs --all-containers


christian-kreuzberger-dtx:
During trying out job-executor I ended up getting this error in Bridge:
```
Job job-executor-service-job-855e19ed-baed-48b3-973e-0e62-1 failed: job job-executor-service-job-855e19ed-baed-48b3-973e-0e62-1 failed. Reason: BackoffLimitExceeded, Message: Job has reached the specified backoff limit
```

Unfortunately this does not tell me why the job has failed, so I had to start troubleshooting on my own.
A very handy troubleshooting step is to look at Kubernetes logs, especially the log of the job or the pod(s).

For this the following steps need to be done:
```console
$ kubectl -n keptn get pods
...
job-executor-service-799fc875c7-hjbjw                           2/2     Running      0          19h
job-executor-service-job-855e19ed-baed-48b3-973e-0e62-1-qdswm   0/1     Init:Error   0          18m
job-executor-service-job-f5f44383-054d-48d0-b219-e2d6-1-p7bbl   0/1     Error        0          19h
job-executor-service-job-ff152cae-1238-4249-b5e3-67fd-1-h2vmc   0/1     Init:Error   0          24m
...
```

Select the pod that you want to inspect, e.g. the most recent one: `job-executor-service-job-855e19ed-baed-48b3-973e-0e62-1-qdswm`

```console
$ kubectl -n keptn logs job-executor-service-job-855e19ed-baed-48b3-973e-0e62-1-qdswm --all-containers
```

And look at the output, e.g.:

```
kubectl -n keptn logs job-executor-service-job-855e19ed-baed-48b3-973e-0e62-1-qdswm -c init-job-executor-service-job-855e19ed-baed-48b3-973e-0e62-1
2021/12/17 10:31:18 Error while copying files: could not find file or directory jmeter/basiccheck.jmx for task Run jmeter smoke tests
```

# Definition of Done

- [ ] Add a troubleshooting section with this as the first troubleshooting step

grabnerandi:
I just ran into the same issue and it took me a while to do some troubleshooting
Would be great to provide those types of messages to the end user as the fix is simple - but - getting to the actual reason is not

christian-kreuzberger-dtx:
This is also addressed using #214, so we don't need to add it to the troubleshooting section.

Issue #125, opened 2021-12-17
Fail with a `.fininished` event with proper message if launching the container fails


christian-kreuzberger-dtx:
WIP

PR #214, opened 2022-04-06
feat: Include logs of all containers in error message

Raffy23:
## This PR
* improves the debugging experience by including all logs of all containers in a pod when reporting back the logs (related to #123)
* adds the failure reason and message / exit code of a container to the error log 

## Solves
* #124

### DIFF
--- a/pkg/k8sutils/pod.go
+++ b/pkg/k8sutils/pod.go
@@ -57,9 +57,12 @@ func (k8s *K8sImpl) GetLogsOfPod(jobName string, namespace string) (string, erro
 				logsOfContainer = fmt.Sprintf("Unable to query logs of container: %s", err.Error())
 			}
 
-			// Build the final logging output for the container
-			logs.WriteString(buildLogOutputForContainer(container, logsOfContainer))
-			logs.WriteString("\n")
+			// If the container did not put out any logs, we skip it entirely to prevent polluting the
+			// log output too much by appending a lot of empty lines for each container
+			if logsOfContainer != "" {
+				logs.WriteString(buildLogOutputForContainer(pod, container, logsOfContainer))
+				logs.WriteString("\n")
+			}
 		}
 	}
 
@@ -140,27 +143,23 @@ func getTerminatedContainersWithStatusOfPod(pod v1.Pod) []containerStatus {
 // format. Depending on the status the output changes slightly (output will be empty of no logs are produced):
 //
 // - Normal output:
-// 		Container <container.name>:
+// 		Container <container.name> of pod <pod.name>:
 //  	<logsOfContainer>
 //
 // - In case of an error:
-// 		Container <container.name> terminated with an error (Reason: <reason> [, Message: <message> |, ExitCode: <code>]):
+// 		Container <container.name> of pod <pod.name> terminated with an error (Reason: <reason> [, Message: <message> |, ExitCode: <code>]):
 // 		<logsOfContainer>
 //
-func buildLogOutputForContainer(container containerStatus, logsOfContainer string) string {
+func buildLogOutputForContainer(pod v1.Pod, container containerStatus, logsOfContainer string) string {
 	var logs strings.Builder
-
-	// If the container did not put out any logs, we skip it entirely to prevent polluting the
-	// log output too much by appending <no logs available for container> for each container
-	if logsOfContainer == "" {
-		return ""
-	}
 
 	// Prepend the container name at the beginning, so we are able to separate logs of different containers
 	// and display a termination error at the beginning, may be more interesting than the logs of the container
 	if container.status.Reason != "Completed" {
 		logs.WriteString("Container ")
 		logs.WriteString(container.name)
+		logs.WriteString(" of pod ")
+		logs.WriteString(pod.Name)
 		logs.WriteString(" terminated with an error (Reason: ")
 		logs.WriteString(container.status.Reason)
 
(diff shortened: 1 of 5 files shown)


pchila:
Just by quickly looking at the code: I am not sure we want to include always the full logs of all containers all the time (think about the successful runs the would include logs from the init container every time) because the output could become polluted.
I would prefer to have the logs of the container that failed (init or main container) in case of failure and only the logs of the main container in case of success... What do you think @christian-kreuzberger-dtx ?

pkg/k8sutils/pod.go
```diff
@@ -3,43 +3,125 @@ package k8sutils
...
-			return "", err
+		// Query all logs from containers that have terminated and therefore already had the chance to
+		// produce logs, otherwise the k8s api will return an error
+		for _, container := range getTerminatedContainersWithStatusOfPod(pod) {
+
+			// Query logs of the current selected container
+			logsOfContainer, err := getLogsOfContainer(k8s, pod, namespace, container.name)
+			if err != nil {
+				return "", err
+			}
```
christian-kreuzberger-dtx, line 35: Returning here means we might lose existing `logs` of previous (terminated) containers. I'd rather add a message to `logs` with the error message contained in `err`.

pkg/k8sutils/pod.go
```diff
@@ -3,43 +3,125 @@ package k8sutils
...
+	buf := new(bytes.Buffer)
+	_, err = io.Copy(buf, podLogs)
+	if err != nil {
+		return "", err
+	}
+
+	return buf.String(), nil
+}
+
+func getTerminatedContainersWithStatusOfPod(pod v1.Pod) []containerStatus {
```
christian-kreuzberger-dtx, line 105: Please add a public comment as well as inline comments to this function

pkg/k8sutils/pod.go
```diff
@@ -3,43 +3,125 @@ package k8sutils
...
+	var containerStatusList []containerStatus
+
+	for index, initContainer := range pod.Spec.InitContainers {
+		if pod.Status.InitContainerStatuses[index].State.Terminated != nil {
```
christian-kreuzberger-dtx, line 109: Can we also look at "pending" containers, containers with "Image Pull Back Off", etc...? Or containers that have never properly started? (same question for the line below with `pod.Spec.Containers`)

pkg/k8sutils/pod.go
```diff
@@ -3,43 +3,125 @@ package k8sutils
...
+
+			// Finally, append the actual logs of the container or a default message to the log
+			if logsOfContainer != "" {
+				logs.WriteString(logsOfContainer)
+				logs.WriteString("\n")
+			} else {
+				logs.WriteString("<no logs available for container>\n")
+			}
+
+			logs.WriteString("\n")
```
christian-kreuzberger-dtx, line 72: I appreciate the efforts taken here, it will make troubleshooting a lot easier. But I'd prefer the "templating stuff" to be put into a separate function, if possible.

christian-kreuzberger-dtx:
@pchila, I believe this is exactly how @Raffy23 implemented it! (Or did he force push some changes :D?)

christian-kreuzberger-dtx (changes requested):
After some deeper review, here are my change requests:


* Do not output initcontainer logs UNLESS initcontainer or any other container fails (if initcontainer succeeds but the 2nd container fails, we need both log outputs)
* Always output container logs of non-initcontainers
* Ensure minimum interference with existing output, e.g.:
**previous output**
```
Job job-executor-service-job-613f1259-f94f-4948-a3c0-7c97-1 finished successfully!

Logs:
...

```
* Please keep the container name though, that's super helpful

pkg/k8sutils/pod.go
Raffy23: I can change the implementation such that it gathers the status of terminated and pending containers, but since we only call this function to gather the logs of jobs that have been marked as finished, no other container states should be possible. I think that the status of pending containers ("Image Pull Backoff") will only be visible in the events.

pchila (approved):
LGTM

christian-kreuzberger-dtx (approved):
It's not solving #125 (still gives me `Task 'Run helm' failed: max poll count reached for job job-executor-service-job-18a4f937-0918-412f-a38a-bdeb-1. Timing out after 5m1.048032241s`), The reason behind is that the pod is not spawning at all :D 
Other then that, it nicely solves #124 and it provides the logs that we needed in a readable and structured format.
LGTM

christian-kreuzberger-dtx:
Log Example:
```
Task 'Run helm' failed: job job-executor-service-job-cba897ca-b3d1-4840-8407-aa77-1 failed. Reason: BackoffLimitExceeded, Message: Job has reached the specified backoff limit

Logs: 
Container init-job-executor-service-job-cba897ca-b3d1-4840-8407-aa77-1 of pod job-executor-service-job-cba897ca-b3d1-4840-8407-aa77-1-fthxr:
2022/05/05 11:53:25 successfully moved file /charts/helloservice.tgz to /keptn/charts/helloservice.tgz


Container job-executor-service-job-cba897ca-b3d1-4840-8407-aa77-1 of pod job-executor-service-job-cba897ca-b3d1-4840-8407-aa77-1-fthxr terminated with an error (Reason: Error, ExitCode: 1):
Release "helloservice" does not exist. Installing it now.
Error: rendered manifests contain a resource that already exists. Unable to continue with install: Service "helloservice" in namespace "podtato-head-production" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "helloservice": current value is "podtato-head-production-helloservice"
```

PR #218, opened 2022-04-08
docs: Provide incompatibility warning for Keptn 0.14.x

christian-kreuzberger-dtx:
Installing Job-Executor-Service with Keptn 0.14.x should work, but it requires us to use "remote execution-plane" setup (which is the default in the current master branch).


pchila (approved):
LGTM

PR #221, opened 2022-04-11
feat: Job security context

Raffy23:
## This PR
* introduces a configurable security context for job workloads


FEATURES.md
```diff
@@ -422,6 +423,24 @@ tasks:
...
+  - name: "Run as different user"
+    securityContext:
+      runAsUser: 7000
+      runAsGroup: 9000
+    ...
+  - name: "Allow modifications to root FS"
+    securityContext:
+      readOnlyRootFilesystem: false
+    ...
+```
```
christian-kreuzberger-dtx, line 442: Please extend both parts with image and command, whereas with the readOnlyRootfilesystem false the command could be an `echo something > test.txt`

cmd/job-executor-service/main.go
```diff
@@ -48,13 +46,15 @@ type envConfig struct {
 	// Default resource requests memory for job and init container
 	DefaultResourceRequestsMemory string `envconfig:"DEFAULT_RESOURCE_REQUESTS_MEMORY"`
 	// Respond with .finished event if no configuration found
-	AlwaysSendFinishedEvent string `envconfig:"ALWAYS_SEND_FINISHED_EVENT"`
+	AlwaysSendFinishedEvent bool `envconfig:"ALWAYS_SEND_FINISHED_EVENT"`
```
christian-kreuzberger-dtx, line 49: I'm wondering if there was a reason why we used `string` here :D 
do we know that `bool` works reliably with envconfig?

cmd/job-executor-service/main.go
```diff
@@ -64,6 +64,25 @@ const ServiceName = "job-executor-service"
...
+ * Parses a Keptn Cloud Event payload (data attribute)
+ */
+func parseKeptnCloudEventPayload(event cloudevents.Event, data interface{}) error {
+	err := event.DataAs(data)
+	if err != nil {
+		log.Fatalf("Got Data Error: %s", err.Error())
+		return err
+	}
+	return nil
+}
```
christian-kreuzberger-dtx, line 84: what's this doing here? Looks like an oversight from a merge?

pkg/utils/securitycontext.go
```diff
@@ -0,0 +1,211 @@
...
+	"encoding/json"
+	"errors"
+	v1 "k8s.io/api/core/v1"
+	"keptn-contrib/job-executor-service/pkg/config"
+	"log"
+	"os"
+)
+
+// jobSecurityContextFilePath describes the path of the security config file that is defined in the deployment.yaml
+const jobSecurityContextFilePath = "/config/job-defaultSecurityContext.json"
```
pchila, line 13: it would be better if these filenames were configurable from the outside of the package (those could easily be constants in main for example or maybe coming from a config file or env var)

pkg/utils/securitycontext.go
```diff
@@ -0,0 +1,211 @@
...
+
+// podSecurityContextFilePath describes the path of the pod security config file that is defined in the deployment.yaml
+const podSecurityContextFilePath = "/config/job-podSecurityContext.json"
+
+// ReadDefaultJobSecurityContext reads the JSON file defined in jobSecurityContextFilePath and parses the output
+// into a valid v1.SecurityContext data structure
+func ReadDefaultJobSecurityContext() (*v1.SecurityContext, error) {
+	jobSecurityContextBytes, err := os.ReadFile(jobSecurityContextFilePath)
+	if err != nil {
+		return nil, err
```
pchila, line 23: make a distinction about the cause of error... This is a problem reading a specific file so for example
```suggestion
		return nil, fmt.Errorf("error reading security context config file %s: %w" jobSecurityContextFilePath, err)
```

pkg/utils/securitycontext.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	var defaultJobSecurityContext v1.SecurityContext
+	err = json.Unmarshal(jobSecurityContextBytes, &defaultJobSecurityContext)
+	if err != nil {
+		return nil, err
```
pchila, line 29: This is a problem with the content of the file itself (we managed to read), so
```suggestion
		return nil, fmt.Errorf("invalid security context configuration format: %w", err)
```

pkg/utils/securitycontext.go
```diff
@@ -0,0 +1,211 @@
...
+
+	return &defaultJobSecurityContext, nil
+}
+
+// ReadDefaultPodSecurityContext reads the JSON file defined in podSecurityContextFilePath and parses the output
+// into a valid v1.PodSecurityContext data structure
+func ReadDefaultPodSecurityContext() (*v1.PodSecurityContext, error) {
+	podSecurityContextBytes, err := os.ReadFile(podSecurityContextFilePath)
+	if err != nil {
+		return nil, err
```
pchila, line 40: same as before

pkg/utils/securitycontext.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	var defaultPodSecurityContext v1.PodSecurityContext
+	err = json.Unmarshal(podSecurityContextBytes, &defaultPodSecurityContext)
+	if err != nil {
+		return nil, err
```
pchila, line 46: here too ;)

pkg/utils/securitycontext.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	return &defaultPodSecurityContext, nil
+}
+
+// BuildSecurityContext builds the final v1.SecurityContext from the SecurityContext defined in the job configuration
+// and the default job security context. The configs are merged in such a way that the configuration in the task will
+// overwrite each entry that is defined.
+func BuildSecurityContext(defaultSecurityContext *v1.SecurityContext, securityContext config.SecurityContext) *v1.SecurityContext {
```
pchila, line 55: here we could use a trick: serializing to JSON and deserializing into the cloned Security Context which would override the values (need to be tested properly). It would be worth trying to avoid the tedious copying of every field

cmd/job-executor-service-lint/main.go
```diff
@@ -19,10 +20,15 @@ func main() {
 		log.Fatalf("could not read job config %v: %v", jobConfigName, err)
 	}
 
-	_, err = config.NewConfig(jobConfig)
+	conf, err := config.NewConfig(jobConfig)
 	if err != nil {
 		log.Fatalf("error parsing %v: %v", string(jobConfig), err)
 	}
 
+	err = utils.VerifySecurityConfiguration(conf, true)
```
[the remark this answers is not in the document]
Raffy23: That's only for the linter, the job-executor-service will query this flag from the environment variable.
I will introduce some command line parsing flags to handle this case correctly.

cmd/job-executor-service-lint/main.go
```diff
@@ -1,15 +1,22 @@
...
 	"keptn-contrib/job-executor-service/pkg/config"
 	"keptn-contrib/job-executor-service/pkg/utils"
 	"log"
-	"os"
 )
 
 func main() {
-	args := os.Args[1:]
+
+	// Parse the allowPrivilegedJobs flag that can be changed to match the behavior of the job-executor-service
```
pchila, line 13: For completeness we should add info about this flag in the README, up to you if you wanna put it in this PR or create a separate issue for docs

pchila (dismissed):

pchila (approved):

PR #222, opened 2022-04-12
test: improve code coverage

pchila:
Refactored `config` and `eventhandler` packages to make them testable and added more unit tests to improve coverage.
Refactored `k8sutils` as well for testability

Overall coverage is now 83.6% with no package below 80% (excluding `fake` packages and `cmd` directories)

image


pkg/eventhandler/eventhandlers_test.go
```diff
@@ -0,0 +1,701 @@
...
+	if reflectedValue.Kind() == reflect.Ptr || reflectedValue.Kind() == reflect.Interface {
+		// if it's a pointer or an interface dereference to get to the struct
+		reflectedValue = reflectedValue.Elem()
+	}
+
+	if am.name != reflectedValue.FieldByName("Name").String() {
+		return false
+	}
+
+	return true
```
Raffy23, line 294: I think we can simplify this to `return am.name == reflectedValue.FieldByName("Name").String()`

Raffy23 (approved):
LGTM

PR #227, opened 2022-04-27
feat: Add Keptn auto-detection

Raffy23:
## This PR
- Adds an auto-detection flag to the JES installation that is able to detect a single Keptn instance and extract the API token from it

### Fixes
- #209

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ In order to install the *job-executor-service* on the remote execution plane, so
 * `remoteControlPlane.api.token` - Keptn API Token (can be obtained from Bridge)
 
 **Example**
+
 ```bash
 KEPTN_API_PROTOCOL=http # or https
 KEPTN_API_HOST=<INSERT-YOUR-HOSTNAME-HERE> # e.g., 1.2.3.4.nip.io
@@ -81,6 +82,22 @@ Please replace `<VERSION>` with the actual version you want to install from the 
 [GitHub releases page](https://github.com/keptn-contrib/job-executor-service/releases).
 
 To verify that everything works you can visit Bridge, select a project, go to Uniform, and verify that `job-executor-service`  is registered as "remote execution plane" with the correct version and event type.
+
+**Example (with auto-detection)**
+
+For easier installation of the *job-executor-service* a single Keptn installation can be auto-detected. This only
+works if no `remoteControlPlane.api.token`, `remoteControlPlane.api.protocol` or `remoteControlPlane.api.hostname` is provided.
+If multiple Keptn installations are present, the `remoteControlPlane.autoDetect.namespace` must be set to the desired Keptn instance.
+The auto-detection feature can be enabled by setting `remoteControlPlane.autoDetect.enabled` to `true`.
+
+```bash
+TASK_SUBSCRIPTION=sh.keptn.event.remote-task.triggered
+
+helm upgrade --install --create-namespace -n <NAMESPACE> \
+  job-executor-service https://github.com/keptn-contrib/job-executor-service/releases/download/<VERSION>/job-executor-service-<VERSION>.tgz \
+ --set remoteControlPlane.autoDetect.enabled=true,remoteControlPlane.topicSubscription=${TASK_SUBSCRIPTION},remoteControlPlane.api.token="",remoteControlPlane.api.hostname="",remoteControlPlane.api.protocol=""
+```
+
 
 ### Update API Token on Remote Execution-Plane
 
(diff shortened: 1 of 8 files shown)


chart/templates/tests/test-api-connection.yaml
```diff
@@ -1,19 +1,28 @@
...
 kind: Pod
 metadata:
   name: "{{ include "job-executor-service.fullname" . }}-test-api-connection"
   labels:
-  {{- include "job-executor-service.labels" . | nindent 4 }}
+    {{- include "job-executor-service.labels" . | nindent 4 }}
   annotations:
     "helm.sh/hook": test
 spec:
+  restartPolicy: Never
```
christian-kreuzberger-dtx, line 10: Wondering why we need this (or better: why we haven't needed this before ^^)?

README.md
```diff
@@ -81,6 +82,22 @@ Please replace `<VERSION>` with the actual version you want to install from the
 
 To verify that everything works you can visit Bridge, select a project, go to Uniform, and verify that `job-executor-service`  is registered as "remote execution plane" with the correct version and event type.
 
+**Example (with auto-detection)**
+
+For easier installation of the *job-executor-service* a single Keptn installation can be auto-detected. This only
+works if no `remoteControlPlane.api.token`, `remoteControlPlane.api.protocol` or `remoteControlPlane.api.hostname` is provided.
```
christian-kreuzberger-dtx, line 88: Maybe rephrase this to "if ... are not provided, e.g., by setting them to an empty string"

README.md
```diff
@@ -81,6 +82,22 @@ Please replace `<VERSION>` with the actual version you want to install from the
...
+If multiple Keptn installations are present, the `remoteControlPlane.autoDetect.namespace` must be set to the desired Keptn instance.
```
christian-kreuzberger-dtx, line 89: If you leave it empty, it defaults to `keptn`?

christian-kreuzberger-dtx:
Just tried this locally using `skaffold`, k3d, and a fresh Keptn 0.13.x installation, and got this error:

```
Helm release job-executor-service not installed. Installing...
Error: INSTALLATION FAILED: template: job-executor-service/templates/secrets.yaml:7:20: executing "job-executor-service/templates/secrets.yaml" at <include "job-executor-service.remote-control-plane.token" .>: error calling include: template: job-executor-service/templates/_helpers.tpl:118:34: executing "job-executor-service.remote-control-plane.token" at <include "job-executor-service.remote-control-plane.namespace" .>: error calling include: template: job-executor-service/templates/_helpers.tpl:100:111: executing "job-executor-service.remote-control-plane.namespace" at <"app.kubernetes.io/part-of">: invalid value; expected string
deploying "job-executor-service": install: exit status 1
```

chart/templates/tests/test-api-connection.yaml
Raffy23: It was there before but on a different position, it seems like re ordering the yaml elements prepended it to the top

README.md
Raffy23: Empty means scan all existing namespaces, a namespace means just have a look at that one.

chart/templates/tests/test-api-connection.yaml
christian-kreuzberger-dtx: Totally overlooked that, thanks!

christian-kreuzberger-dtx:
Just tried it on a fresh installation with k3d, and got this:
```
Error: INSTALLATION FAILED: create: failed to create: namespaces "keptn-jes" not found
```

adding
```
    flags:
      install: ["--create-namespace"]
```

fixed it for me, and it's working :+1:

chart/templates/_helpers.tpl
```diff
@@ -72,6 +72,58 @@ Create the name of the job service account to use
...
+
+    {{- if eq (len $detectedKeptnNamespaces) 0 }}
+        {{- fail "Unable to detect Kepn in the kubernetes cluster!" }}
+    {{- end }}
+    {{- if gt (len $detectedKeptnNamespaces) 1 }}
+        {{- fail "Detected more than one Keptn installation!" }}
+    {{- end }}
+
+    {{- index $detectedKeptnNamespaces 0 }}
+    {{- end }}
```
pchila, line 113: I would get rid of using the `.Values.remoteControlPlane.autoDetect.namespace` and always do a lookup, since that value is supposed to look into a particular namespace but I would like to still check for the api-gateway-nginx

```suggestion
    {{- $detectedKeptnApiGateways := list }}

    {{- /* Find api-gateway-nginx service, which is used as keptn api gatway */ -}}

    {{- $services := lookup "v1" "Service" (.Values.remoteControlPlane.autoDetect.namespace | default "") "" }}

    {{- range $index, $srv := $services.items }}
        {{- if and (eq "api-gateway-nginx" $srv.name ) (hasPrefix "keptn-" ( get $srv.metadata.labels "app.kubernetes.io/part-of" )) }}
            {{- $detectedKeptnApiGateways = append $detectedKeptnApiGateways $srv }}
        {{- end }}
    {{- end }}

    {{- if eq (len $detectedKeptnApiGateways) 0 }}
        {{- fail "Unable to detect Kepn in the kubernetes cluster!" }}
    {{- end }}
    {{- if gt (len $detectedKeptnApiGateways) 1 }}
        {{- fail (printf "Detected more than one Keptn installation: %+v" $detectedKeptnApiGateways) }}
    {{- end }}

    {{- (index $detectedKeptnApiGateways 0).metadata.namespace }}
```

pchila (approved):
LGTM

PR #229, opened 2022-04-27
fix: Job cleanup integration test

Raffy23:
## This PR
* implements a fix such that JES reports the status correctly for jobs that will be collected instantly
* enables the Job cleanup test with TTL 0


christian-kreuzberger-dtx (changes requested):
Let's not merge this for now, there might be a better solution (e.g., limiting TTL to values >= 60).

PR #240, opened 2022-05-09
feat: Add job labels

Raffy23:
## This PR
* adds generated and user defined labels to Kubernetes jobs

### DIFF
--- a/chart/templates/configmap.yaml
+++ b/chart/templates/configmap.yaml
@@ -14,5 +14,8 @@ data:
   init_container_configuration_endpoint: {{ include "job-executor-service.remote-control-plane.configuration-endpoint" . }}
   default_job_service_account: "{{ include "job-executor-service.jobConfig.serviceAccountName" . }}"
   allow_privileged_jobs: "{{ .Values.jobConfig.allowPrivilegedJobs | default "false" }}"
-  additional_job_labels: {{ .Values.jobConfig.labels | toJson | quote }}
+  additional_job_labels: |
+    {{- with .Values.jobConfig.labels }}
+      {{- toYaml . | nindent 6 }}
+    {{- end }}
   task_deadline_seconds: {{ .Values.jobConfig.taskDeadlineSeconds | default 0 | quote}}
--- a/chart/templates/configmap.yaml
+++ b/chart/templates/configmap.yaml
@@ -14,4 +14,5 @@ data:
   init_container_configuration_endpoint: {{ include "job-executor-service.remote-control-plane.configuration-endpoint" . }}
   default_job_service_account: "{{ include "job-executor-service.jobConfig.serviceAccountName" . }}"
   allow_privileged_jobs: "{{ .Values.jobConfig.allowPrivilegedJobs | default "false" }}"
+  additional_job_labels: {{ .Values.jobConfig.labels | toJson | quote }}
   task_deadline_seconds: {{ .Values.jobConfig.taskDeadlineSeconds | default 0 | quote}}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -131,9 +131,9 @@ func NewConfig(yamlContent []byte) (*Config, error) {
 }
 
 // IsEventMatch indicated whether a given event matches the config
-func (c *Config) IsEventMatch(eventType string, jsonEventData interface{}) (bool, *Action) {
+func (c *Config) IsEventMatch(eventType string, jsonEventData interface{}) (bool, *Action, int) {
 
-	for _, action := range c.Actions {
+	for index, action := range c.Actions {
 		for _, event := range action.Events {
 			// does the event type match with regex?
 			matched, _ := regexp.MatchString(event.Name, eventType)
@@ -141,7 +141,7 @@ func (c *Config) IsEventMatch(eventType string, jsonEventData interface{}) (bool
 
 				// no JSONPath specified, just match event type
 				if event.JSONPath.Property == "" {
-					return true, &action
+					return true, &action, index
 				}
 
 				value, err := jsonpath.Get(event.JSONPath.Property, jsonEventData)
@@ -150,12 +150,12 @@ func (c *Config) IsEventMatch(eventType string, jsonEventData interface{}) (bool
 				}
 
 				if value == event.JSONPath.Match {
-					return true, &action
+					return true, &action, index
 				}
 			}
 		}
 	}
-	return false, nil
+	return false, nil, -1
 }
 
 // FindActionByName searches for a given Action by a provided name within the config
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -131,31 +131,41 @@ func NewConfig(yamlContent []byte) (*Config, error) {
 }
 
 // IsEventMatch indicated whether a given event matches the config
-func (c *Config) IsEventMatch(eventType string, jsonEventData interface{}) (bool, *Action) {
+func (c *Config) IsEventMatch(eventType string, jsonEventData interface{}) bool {
 
 	for _, action := range c.Actions {
-		for _, event := range action.Events {
-			// does the event type match with regex?
-			matched, _ := regexp.MatchString(event.Name, eventType)
-			if matched {
+		if action.IsEventMatch(eventType, jsonEventData) {
+			return true
+		}
+	}
+	return false
+}
 
-				// no JSONPath specified, just match event type
-				if event.JSONPath.Property == "" {
-					return true, &action
-				}
+// IsEventMatch indicated whether a given event matches the action
+func (a *Action) IsEventMatch(eventType string, jsonEventData interface{}) bool {
 
-				value, err := jsonpath.Get(event.JSONPath.Property, jsonEventData)
-				if err != nil {
-					continue
-				}
+	for _, event := range a.Events {
+		// does the event type match with regex?
+		matched, _ := regexp.MatchString(event.Name, eventType)
+		if matched {
 
-				if value == event.JSONPath.Match {
-					return true, &action
-				}
+			// no JSONPath specified, just match event type
+			if event.JSONPath.Property == "" {
+				return true
+			}
+
+			value, err := jsonpath.Get(event.JSONPath.Property, jsonEventData)
+			if err != nil {
+				continue
+			}
+
+			if value == event.JSONPath.Match {
+				return true
 			}
 		}
 	}
-	return false, nil
+
+	return false
 }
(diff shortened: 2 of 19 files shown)


pkg/config/config.go
```diff
@@ -150,12 +150,12 @@ func (c *Config) IsEventMatch(eventType string, jsonEventData interface{}) (bool
...
 
 				if value == event.JSONPath.Match {
-					return true, &action
+					return true, &action, index
 				}
 			}
 		}
 	}
-	return false, nil
+	return false, nil, -1
```
pchila, line 158: What does -1 mean? If we use magic number and assign value to them please document them using a constant

pkg/eventhandler/eventhandlers.go
```diff
@@ -173,6 +175,9 @@ func (eh *EventHandler) startK8sJob(action *config.Action, jsonEventData interfa
 		}
 	}
 
+	// The action ID is generated in such a way that it is unique to the current context of the event
+	actionID := fmt.Sprintf("%d", actionIndex)
```
pchila, line 179: We have the action name, why are we using the position in a list as ID ?
If it's to distinguish actions with the same name we should probably enforce the uniqueness of names or compose some other ID based on it instead of using an index which may change from one version of job config to another...

pkg/k8sutils/job.go
```diff
@@ -442,3 +458,44 @@ func (k8s *K8sImpl) generateEnvFromSecret(env config.Env, namespace string) ([]v
...
+
+	splitJobName := strings.Split(jobName, "-")
+	jobIndex, err := strconv.Atoi(splitJobName[len(splitJobName)-1])
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse task index")
+	}
+
+	// Job index in the name starts at 1
+	jobIndex = jobIndex - 1
+
```
pchila, line 492: We are introducing a dependency on a string that has been constructed in another package, so we are introducing a dependency on the format of such string between packages. 
We have the task index used for the last part of the job name, which was an int, then we transform it into a string, then we concatenate it, then we split it and re-parse it here.

If we need to manipulate information when creating k8s job we should have all the related code in one package and the inputs should be passed into the package with their original types so that all tightly coupled code is an implementation detail inside one package instead of being a shared responsibility (and dependency!) across packages

chart/templates/configmap.yaml
```diff
@@ -14,3 +14,4 @@ data:
   init_container_configuration_endpoint: {{ include "job-executor-service.remote-control-plane.configuration-endpoint" . }}
   default_job_service_account: "{{ include "job-executor-service.jobConfig.serviceAccountName" . }}"
   allow_privileged_jobs: "{{ .Values.jobConfig.allowPrivilegedJobs | default "false" }}"
+  additional_job_labels: {{ .Values.jobConfig.labels | toJson | quote }}
```
pchila, line 17: why do we convert the labels (which I am assuming is a map[string]string) to a json string ?

pkg/eventhandler/fake/eventhandlers_mock.go
```diff
@@ -180,31 +180,17 @@ func (mr *MockK8sMockRecorder) ConnectToCluster() *gomock.Call {
...
 }
 
 // CreateK8sJob indicates an expected call of CreateK8sJob.
-func (mr *MockK8sMockRecorder) CreateK8sJob(arg0, arg1, arg2, arg3, arg4, arg5, arg6 interface{}) *gomock.Call {
+func (mr *MockK8sMockRecorder) CreateK8sJob(arg0, arg1, arg2, arg3, arg4, arg5, arg6, arg7 interface{}) *gomock.Call {
 	mr.mock.ctrl.T.Helper()
-	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "CreateK8sJob", reflect.TypeOf((*MockK8s)(nil).CreateK8sJob), arg0, arg1, arg2, arg3, arg4, arg5, arg6)
-}
-
-// ExistsServiceAccount mocks base method.
```
pchila, line 196: It seems we lost this function from the interface, yet I don't see any related changes in this PR, why is this change here ?

pkg/k8sutils/job_test.go
```diff
@@ -817,3 +828,209 @@ func createK8sSecretObj(name string, namespace string, data map[string][]byte) *
...
+func TestCreateK8sJobContainsCorrectLabels(t *testing.T) {
+	k8sClientSet := k8sfake.NewSimpleClientset()
+	k8s := K8sImpl{clientset: k8sClientSet}
+
+	// Prepend reactor that we are able to assert metadata labels
+	k8sClientSet.PrependReactor(
+		"create", "jobs", func(action k8stesting.Action) (handled bool, ret runtime.Object, err error) {
+			job := action.(k8stesting.CreateAction).GetObject().(*v1.Job)
+
+			errors := validation.ValidateLabels(job.ObjectMeta.Labels, &field.Path{})
```
pchila, line 841: why not doing a assert on the labels here directly ?

test/events/test.triggered.with-gitcommitid.json
```diff
@@ -0,0 +1,33 @@
+{
```
pchila, line 1: where is this new file used ?

pkg/eventhandler/eventhandlers.go
Raffy23: Currently we actually don't try to distinguish actions at all, we just take the first one that matches the event, which also means that the name is currently not required to be unique. If we can make the action name unique that would be a better way to identify the action.

pkg/eventhandler/fake/eventhandlers_mock.go
Raffy23: I think the function was only present in the mocking code or the implementation of #229. I think it somehow slipped into a previous PR.

chart/templates/configmap.yaml
Raffy23: I'm not sure if a yaml map can be used without any problems as environment variable, but I will try to avoid converting

pkg/eventhandler/eventhandlers.go
pchila: @christian-kreuzberger-dtx  What do you think about enforcing unique names that can be used as label values for the actions? How much of a breaking change would it be ?

test/events/test.triggered.with-gitcommitid.json
Raffy23: The file is used in the newly added unit tests.

pkg/k8sutils/job_test.go
```diff
@@ -817,3 +867,188 @@ func createK8sSecretObj(name string, namespace string, data map[string][]byte) *
...
+			require.NoError(t, err)
+
+			expectedLabels := map[string]string{
+				"app.kubernetes.io/managed-by": "job-executor-service",
+				"keptn.sh/context":             test.event["shkeptncontext"].(string),
+				"keptn.sh/ceid":                test.event["id"].(string),
+				"keptn.sh/commitid":            "",
+				"keptn.sh/jes-action":          test.expectedActionName,
+				"keptn.sh/jes-task":            test.expectedTaskName,
+				"keptn.sh/job-confighash":      "",
```
pchila, line 971: just for consistency: shouldn't this label start with `keptn.sh/jes-*`?

```suggestion
				"keptn.sh/jes-job-confighash":      "",
```

pkg/k8sutils/job_test.go
```diff
@@ -817,3 +867,188 @@ func createK8sSecretObj(name string, namespace string, data map[string][]byte) *
...
+			)
+			require.NoError(t, err)
+
+			job, err := k8sClientSet.BatchV1().Jobs(namespace).Get(context.TODO(), jobName, metav1.GetOptions{})
+			require.NoError(t, err)
+
+			expectedLabels := map[string]string{
+				"app.kubernetes.io/managed-by": "job-executor-service",
+				"keptn.sh/context":             test.event["shkeptncontext"].(string),
+				"keptn.sh/ceid":                test.event["id"].(string),
```
pchila, line 967: Is this the cloud event id? we can avoid the acronym maybe
```suggestion
				"keptn.sh/event-id":                test.event["id"].(string),
```

pchila (approved):
Just a couple of hints for consistency and readability of labels, otherwise it looks good.

pchila (dismissed):
LGTM

pkg/config/reader.go
```diff
@@ -21,19 +22,25 @@ type JobConfigReader struct {
...
 // error) with the original error correctly wrapped in the local one
-func (jcr *JobConfigReader) GetJobConfig() (*Config, error) {
+func (jcr *JobConfigReader) GetJobConfig() (*Config, string, error) {
 	resource, err := jcr.Keptn.GetKeptnResource(jobConfigResourceName)
 	if err != nil {
-		return nil, fmt.Errorf("error retrieving job config: %w", err)
+		return nil, "", fmt.Errorf("error retrieving job config: %w", err)
 	}
 
+	hasher := sha1.New()
```
christian-kreuzberger-dtx, line 34: could/should we use `sha256` instead of the (AFAIK outdated and insecure) `sha1`?

Raffy23: I have the feeling that sha256 is too long to be used as a label value. In this case we usually don't care about security, but just that the configuration file didn't change. But I can check if sha256 or sha224 works.

pkg/eventhandler/eventhandlers.go
```diff
@@ -115,26 +115,32 @@ func (eh *EventHandler) HandleEvent() error {
...
-		"Match found for event %s of type %s. Starting k8s job to run action '%s'", eh.Keptn.CloudEvent.Context.GetID(),
-		eh.Keptn.CloudEvent.Type(), action.Name,
-	)
+	// For each action that matches the given event type we execute all containing tasks:
+	for actionIndex, action := range configuration.Actions {
+		if action.IsEventMatch(eh.Keptn.CloudEvent.Type(), eventAsInterface) {
+			log.Printf(
+				"Match found for event %s of type %s. Starting k8s job to run action '%s'", eh.Keptn.CloudEvent.Context.GetID(),
+				eh.Keptn.CloudEvent.Type(), action.Name,
+			)
```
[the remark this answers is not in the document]
Raffy23: I haven't checked the issue, but it might solve it. The change was more or less necessary to get the index of the targeted action without producing too hacky code.

christian-kreuzberger-dtx (approved):
LGTM

PR #241, opened 2022-05-09
feat: Enforce minimum job TTL value

Raffy23:
## This PR
* enforces a minimal TTL value of 60s
* enables the TTL integration tests

### DIFF
--- a/pkg/config/reader.go
+++ b/pkg/config/reader.go
@@ -7,8 +7,6 @@ import (
 
 const jobConfigResourceName = "job/config.yaml"
 
-const minTTLSecondsAfterFinished = int32(60)
-
 //go:generate mockgen -destination=fake/reader_mock.go -package=fake . KeptnResourceService
 
 // KeptnResourceService defines the contract used by JobConfigReader to retrieve a resource from keptn (using project,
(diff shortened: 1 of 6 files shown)


pkg/config/reader.go
```diff
@@ -35,5 +37,23 @@ func (jcr *JobConfigReader) GetJobConfig() (*Config, error) {
 		log.Printf("The config was: %s", string(resource))
 		return nil, fmt.Errorf("error parsing job configuration: %w", err)
 	}
+
+	// Validate / Correct the Job cleanup TTL to 60s to avoid a job cleanup before the
```
pchila, line 41: As a matter of responsibility, I would have put this snippet in the k8sutils package and verified that we enforce the minimum TTL when creating the job on k8s (the jobconfig reader has the responsibility of retrieving and parsing the config yaml, while the minimum ttl is needed by the k8s polling mechanism in k8sutils).

If the implementation for checking the job status changes we now have 2 packages impacted instead of only one.

christian-kreuzberger-dtx (approved):
LGTM

pchila (approved):
LGTM

PR #248, opened 2022-05-16
chore: set helm chart version to 0.0.0-dev

christian-kreuzberger-dtx:
Set helm chart version to 0.0.0-dev (was 0.1.4 before).

This helps when deploying the chart using skaffold.

Note: The actual version is set as part of the build-helm-chart workflow:

image


pchila (approved):
LGTM

Issue #244, opened 2022-05-12
Refactor `K8sImpl` structure to simplify the interaction with kubernetes


Raffy23:
At the moment the interaction between the job-executor-service and kubernetes can be improved by refactoring the following things: 

* The job-executor-service uses polling in the `AwaitK8sJobDone` function, we should be able to refactor this function to utilize the watch functionality from the kubernetes API and simplify a lot of logic when it comes to handling long running jobs.
* The `K8sImpl` structure can probably contain a lot of information which is passed around with the `JobSettings` structure, as these structures are initialized at the same time and the information in the `JobSettings` struct aren't changed.
* While it is necessary to pass the cloud event data around with the `jsonEventData` parameter, we should invest some time to provide a more structured way of accessing the contents of the cloud event.

With these changes we could not only improve the handling of long running / failed job states, but also generalize the K8s interface such that it gets easier to mock functions for testing.

PR #249, opened 2022-05-16
fix: Add output of failed events to logs

### LINKED ISSUE #234: Specifying non-existing Service Account leads to job not starting
This was testedwith jes pre-release 0.2.0-next.0.

When specifying serviceAccount of a job, e.g.,

```
  - name: "Deploy using helm"
    events:
      - name: "sh.keptn.event.je-deployment.triggered"
    tasks:
      - name: "Run helm"
        files:
          - /charts
        env:
          - name: IMAGE
            value: "$.data.configurationChange.values.image"
            valueFrom: event
        image: "alpine/helm:3.7.2"
        serviceAccount: "jes-deploy-using-helm"
        cmd: ["helm"]
```

without `jes-deploy-using-helm` existing in the current namespace, the job is not starting (and eventually times out).

# Expected Behaviour

There should be an error sent back to Keptn (either as a .finished event with status fail/error, or a dedicated error log).

### LINKED ISSUE #235: Job that cannot start pod -> orphaned
When Job Executor Service creates a k8s Job that cannot spawn pod (if the wrong serviceAccount for the task is specified for example) the sequence fails because of timeout but no logs can be retrieved as there are no pods to fetch them from.
Furthermore the created job is not collected by the k8s TTL controller as it never finishes, so it will keep trying to spawn pods long after Job Executor Service gave up on it (possibly indefinitely if there's a configuration error) so it has to be manually removed by the user.

The Job Executor Service should detect that the job failed to start and add relevant information for the user extracted from the job status/events and explicitly delete the job if it didn't spawn any pods.

### How to reproduce:
Use a job config with a wrong service account:

```
apiVersion: v2
actions:
  - name: "Hello World e2e test"
    events:
      - name: "sh.keptn.event.deployment.triggered"
    tasks:
      - name: "Greet the world"
        image: "alpine"
        serviceAccount: "inexistentServiceAccount"
        cmd:
          - echo
        args:
          - "Hello World"
```

christian-kreuzberger-dtx:
Disclaimer: Thanks to the code provided by @Raffy23 in PR https://github.com/keptn-contrib/job-executor-service/pull/229

Fixes #234 by providing an appropriate error message

image

Unfortunately, we cannot easily fix #235, as it would require us to delete jobs (which we shouldn't do at the moment). I would dedicate the fix of #235 to the refactoring ticket of #244

### DIFF
--- a/pkg/k8sutils/job_test.go
+++ b/pkg/k8sutils/job_test.go
@@ -740,6 +740,88 @@ func TestAwaitK8sJobDoneErrorJobFailed(t *testing.T) {
 	)
 }
 
+func TestAwaitK8sJobDoneNotStarted(t *testing.T) {
+	k8sClientSet := k8sfake.NewSimpleClientset()
+	k8s := K8sImpl{clientset: k8sClientSet}
+
+	jobName := "job-not-started"
+
+	job := v1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: jobName,
+		},
+		Spec: v1.JobSpec{},
+		Status: v1.JobStatus{
+			Conditions: []v1.JobCondition{}, // this job won't have a condition set
+		},
+	}
+	namespace := "job-pain-and-misery-ns"
+	k8sClientSet.BatchV1().Jobs(namespace).Create(
+		context.Background(), &job, metav1.CreateOptions{},
+	)
+
+	err := k8s.AwaitK8sJobDone(jobName, 1*time.Second, 50*time.Millisecond, namespace)
+
+	require.Error(t, err)
+
+	assert.ErrorContains(
+		t, err, "max poll count reached for job",
+	)
+}
+
+func TestGetFailedEventsForJob(t *testing.T) {
+	k8sClientSet := k8sfake.NewSimpleClientset()
+	k8s := K8sImpl{clientset: k8sClientSet}
+
+	jobName := "happy-path-job"
+	namespace := "happy-path-ns"
+	failedEventMessage1 := "Failed to create pod"
+	failedEventMessage2 := "Failed to pull image"
+	successfulEventMessage1 := "Successfully pulled image"
+
+	// job config
+	job := v1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: jobName,
+		},
+		Spec: v1.JobSpec{},
+		Status: v1.JobStatus{
+			Conditions: []v1.JobCondition{}, // this job won't have a condition set
+		},
+	}
+
+	// two failed and one successful event
+	events := []corev1.Event{
+		{ObjectMeta: metav1.ObjectMeta{Name: "event-1"}, TypeMeta: metav1.TypeMeta{Kind: "job"}, InvolvedObject: corev1.ObjectReference{Name: jobName, Kind: "job", Namespace: namespace}, Source: corev1.EventSource{}, Reason: "FailedCreate", Message: failedEventMessage1},
+		{ObjectMeta: metav1.ObjectMeta{Name: "event-2"}, TypeMeta: metav1.TypeMeta{Kind: "job"}, InvolvedObject: corev1.ObjectReference{Name: jobName, Kind: "job", Namespace: namespace}, Reason: "Failed", Message: failedEventMessage2},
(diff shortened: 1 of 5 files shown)


Raffy23:
Shouldn't we also implement a basic unit test for `GetFailedEventsForJob`?

christian-kreuzberger-dtx:
Added two unit tests and rebased `main` branch

pkg/k8sutils/job_test.go
```diff
@@ -740,6 +740,88 @@ func TestAwaitK8sJobDoneErrorJobFailed(t *testing.T) {
...
+			Conditions: []v1.JobCondition{}, // this job won't have a condition set
+		},
+	}
+
+	// two failed and one successful event
+	events := []corev1.Event{
+		{ObjectMeta: metav1.ObjectMeta{Name: "event-1"}, TypeMeta: metav1.TypeMeta{Kind: "job"}, InvolvedObject: corev1.ObjectReference{Name: jobName, Kind: "job", Namespace: namespace}, Source: corev1.EventSource{}, Reason: "FailedCreate", Message: failedEventMessage1},
+		{ObjectMeta: metav1.ObjectMeta{Name: "event-2"}, TypeMeta: metav1.TypeMeta{Kind: "job"}, InvolvedObject: corev1.ObjectReference{Name: jobName, Kind: "job", Namespace: namespace}, Reason: "Failed", Message: failedEventMessage2},
+		{ObjectMeta: metav1.ObjectMeta{Name: "event-3"}, TypeMeta: metav1.TypeMeta{Kind: "job"}, InvolvedObject: corev1.ObjectReference{Name: jobName, Kind: "job", Namespace: namespace}, Reason: "Pulled", Message: successfulEventMessage1},
+	}
```
christian-kreuzberger-dtx, line 798: I'm not 100% sure if what I'm doing here is correct - haven't had much experience with "faking" events in the fake client. Please help me out ^^

pkg/k8sutils/job_test.go
```diff
@@ -740,6 +740,88 @@ func TestAwaitK8sJobDoneErrorJobFailed(t *testing.T) {
...
+		context.Background(), &job, metav1.CreateOptions{},
+	)
+
+	// create events
+	for _, event := range events {
+		_, err := k8sClientSet.CoreV1().Events(namespace).Create(
+			context.Background(), &event, metav1.CreateOptions{},
+		)
+
+		require.NoError(t, err)
```
christian-kreuzberger-dtx, line 811: Just want to check if adding the event works... sometimes it fails (e.g., duplicate name)

pkg/k8sutils/job_test.go
```diff
@@ -740,6 +740,88 @@ func TestAwaitK8sJobDoneErrorJobFailed(t *testing.T) {
...
+	}
+
+	// create job
+	k8sClientSet.BatchV1().Jobs(namespace).Create(
+		context.Background(), &job, metav1.CreateOptions{},
+	)
+
+	// create events
+	for _, event := range events {
+		_, err := k8sClientSet.CoreV1().Events(namespace).Create(
```
christian-kreuzberger-dtx, line 807: Is it possible to add more than one event with a single function call?

pkg/k8sutils/job_test.go
```diff
@@ -740,6 +740,88 @@ func TestAwaitK8sJobDoneErrorJobFailed(t *testing.T) {
...
+	successfulEventMessage1 := "Successfully pulled image"
+
+	// job config
+	job := v1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: jobName,
+		},
+		Spec: v1.JobSpec{},
+		Status: v1.JobStatus{
+			Conditions: []v1.JobCondition{}, // this job won't have a condition set
```
Raffy23, line 789: I'm not sure about the Kubernetes API at this point, but shouldn't the JobStatus be completely empty if the job hangs in at the point of not being able to create a single pod?

Raffy23 (approved):

pkg/k8sutils/job_test.go
Raffy23: I don't think that the k8s API provides a convenient way to create multiple events, but I will let you know if I find a solution.

pkg/k8sutils/job_test.go
christian-kreuzberger-dtx: Using it the same way as in other tests - I believe that's the correct approach

PR #255, opened 2022-05-20
build(deps): Upgrade setup-go action to use go.mod as a version file

abtris:


.github/workflows/CI.yml
```diff
@@ -207,9 +206,9 @@ jobs:
...
     steps:
       - name: Install Go
-        uses: actions/setup-go@v3.0.0
+        uses: actions/setup-go@v3.1.0
         with:
-          go-version: 1.17
+          go-version-file: "go.mod"
 
       - name: Checkout Code
         uses: actions/checkout@v3
```
christian-kreuzberger-dtx, line 214: You need to move the checkout action before the install go action, in order to be able to read go.mod

abtris: Updated.

christian-kreuzberger-dtx (approved):
LGTM

PR #259, opened 2022-05-27
feat: limit job executor service network access

pchila:


docs/FEATURES.md
```diff
@@ -522,6 +522,27 @@ helm upgrade --install --create-namespace -n <NAMESPACE> \
...
+  - k8s apiserver.
+
+This setting should limit the possibility of interaction with the job-executor-service from malicious third parties and
+limit the potential to interfere with other workloads on the same kubernetes cluster.
+
+To install the network-policy specify
+```yaml
+networkPolicy:
+  enabled: true
+```
```
Raffy23, line 543: I think when referencing the installation we should give an example with or explain that appending `--set networkPolicy.enabled=true` works to avoid confusion for people that are not very familiar with helm.

pchila: added the `--set` equivalent in docs :)

Raffy23 (approved):
LGTM

PR #265, opened 2022-06-02
feat: OAuth authentication mode

Raffy23:
## This PR
* adds OAuth to the job executor service
* adds OAuth to the ini-containers

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -39,7 +39,7 @@ remoteControlPlane:
     apiValidateTls: true                     # Defines if the control plane certificate should be validated
     hostname: "api-gateway-nginx.keptn"      # Hostname of the control plane cluster, Port
     authMode: "token"                        # Flag to choose the defined authentication method, valid values are token and oauth
-    token: ""                                # Keptn API Token (leave empty if OAuth should be used)
+    token: ""                                # Keptn API Token
     oauth:                                   #
       clientId: ""                           # The client OAuth ID
       clientSecret: ""                       # The client OAuth Secret
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -38,7 +38,13 @@ remoteControlPlane:
     protocol: "http"                         # Used Protocol (http, https)
     apiValidateTls: true                     # Defines if the control plane certificate should be validated
     hostname: "api-gateway-nginx.keptn"      # Hostname of the control plane cluster, Port
-    token: ""                                # Keptn API Token
+    authMode: "token"                        # Flag to choose the defined authentication method, valid values are token and oauth
+    token: ""                                # Keptn API Token (leave empty if OAuth should be used)
+    oauth:                                   #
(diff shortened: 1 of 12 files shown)


chart/values.yaml
```diff
@@ -38,7 +38,13 @@ remoteControlPlane:
     protocol: "http"                         # Used Protocol (http, https)
     apiValidateTls: true                     # Defines if the control plane certificate should be validated
     hostname: "api-gateway-nginx.keptn"      # Hostname of the control plane cluster, Port
-    token: ""                                # Keptn API Token
+    authMode: "token"                        # Flag to choose the defined authentication method, valid values are token and oauth
+    token: ""                                # Keptn API Token (leave empty if OAuth should be used)
```
pchila, line 42: I think that the comment is no longer relevant...
```suggestion
    token: ""                                # Keptn API Token
```

cmd/job-executor-service-initcontainer/main.go
```diff
@@ -40,12 +56,54 @@ func main() {
...
+	}
+
+	apiOptions := []func(*api.APISet){
+		api.WithScheme(baseURL.Scheme),
+	}
+
+	// Chose the authentication method from the give environment; This can either be token or oauth
+	if env.AuthMode == "token" {
+
+		// We only append a token when needing / having one, no token is needed
```
pchila, line 71: What do you mean exactly with `no token is needed if the init-container lives in the same namespace as keptn` ? I thought authentication would be necessary anyways.... Is this a feature of the `api.APISet` ?

cmd/job-executor-service-initcontainer/main.go
```diff
@@ -40,12 +56,54 @@ func main() {
...
+		// is part of a remote execution plane then a token is needed
+		if env.KeptnAPIToken != "" {
+			apiOptions = append(apiOptions, api.WithAuthToken(env.KeptnAPIToken, "x-token"))
+		}
+
+	} else if env.AuthMode == "oauth" {
+
+		// To avoid stalling the Job for too long we wait at max 10 seconds to query the token endpoint
+		// from the given discovery url. This is the same timeout that is used in the distributor.
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
```
pchila, line 82: Please no magic numbers, define a constant like `OAuthDiscoveryTimeout` or similar...

cmd/job-executor-service-initcontainer/main.go
Raffy23: It's a description that matches the old behavior, if there is no token we would use an unauthenticated resource handler to speak to the configuration service. I'm not sure if this is still possible with newer versions of Keptn and we maybe should just throw an error if there is no token specified ...

pchila (dismissed):
LGTM

pchila (approved):
LGTM :)

PR #267, opened 2022-06-07
feat: verify signed config (DO NOT MERGE)

pchila:
Experiment to try and use [OpenSSH file signing format ](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.sshsig) to validate a job configuration.

Most of the code deals with decoding the ASCII armored signature while the signature validation is taken care of by [crypto/ssh](https://pkg.go.dev/golang.org/x/crypto/ssh) package


pkg/config/signing/ssh/verify.go
```diff
@@ -0,0 +1,191 @@
...
+			publicKey.Type(), keyComment, keyOptions,
+		)
+		verificationErr := publicKey.Verify(signedMessage, signature.signature)
+		if verificationErr == nil {
+			log.Print("Job config signature verification succeeded")
+			// Verification succeeded :)
+			return nil
+		}
+		log.Print("Job config signature verification failed")
+	}
```
christian-kreuzberger-dtx, line 169: This for loop does not make much sense, does it?
Also, we should return some error message at the end if the verification failed

pchila: This for loop tries all the public keys that exist in the file to check if any of those verifies the signature.
If we had principal information on the authorized signers file and the signature itself we could check only the matching public key but that's not available in the interfaces (may be available on the concrete implementations) but that would be a change in format that I don't think is supported in openssh

PR #275, opened 2022-06-13
feat: Upgrade to Keptn 0.14

TannerGabriel:
## This PR

- Upgrades distributor to Keptn 0.14.1
- Upgrades `go-utils` to v0.14
- Updates compatibility matrix

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ running in the Keptn ecosystem:
 |    0.12.2     |                               keptncontrib/job-executor-service:0.1.7                                |       v2       |
 |    0.12.6     |                               keptncontrib/job-executor-service:0.1.8                                |       v2       |
 |    0.13.x     |                               keptncontrib/job-executor-service:0.1.9                                |       v2       |
-|    0.13.x     |                               keptncontrib/job-executor-service:0.1.9                                |       v2       |
+|    0.13.x     |                               keptncontrib/job-executor-service:0.2.0                                |       v2       |
 |    0.14.x     |                               keptncontrib/job-executor-service:0.2.1                                |       v2       |
 
 Please note: Newer Keptn versions might be compatible, but compatibility has not been verified at the time of the release.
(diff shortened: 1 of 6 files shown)


Raffy23 (approved):
LGTM

README.md
```diff
@@ -29,9 +29,14 @@ running in the Keptn ecosystem:
 |    0.12.2     |                               keptncontrib/job-executor-service:0.1.7                                |       v2       |
 |    0.12.6     |                               keptncontrib/job-executor-service:0.1.8                                |       v2       |
 |    0.13.x     |                               keptncontrib/job-executor-service:0.1.9                                |       v2       |
+|    0.13.x     |                               keptncontrib/job-executor-service:0.1.9                                |       v2       |
+|    0.14.x     |                               keptncontrib/job-executor-service:0.2.1                                |       v2       |
 
 Please note: Newer Keptn versions might be compatible, but compatibility has not been verified at the time of the release.
 
+**Note**: Versions compatible with Keptn 0.14.x and newer are not backward compatible due to a change in NATS cluster name
```
pchila, line 37: I would add that this limitation exists only for installation in the same namespace as keptn (the default installation in a separate namespace does not have this constraint)

chart/values.yaml
```diff
@@ -18,7 +18,7 @@ distributor:
   image:
     repository: docker.io/keptn/distributor  # Container Image Name
     pullPolicy: IfNotPresent                 # Kubernetes Image Pull Policy
-    tag: "0.13.4"                             # Container Tag
+    tag: "0.14.1"                             # Container Tag
```
pchila, line 21: Please use at least 0.14.2 by default, keptn 0.14.1 has a well known bug in the aggregation of deploymentURIs and frankly I would try to avoid any reference to that version (even if the distributor is not involved)

skaffold.yaml
```diff
@@ -55,7 +55,7 @@ deploy:
         overrides:
           distributor:
             image:
-              tag: 0.13.4
+              tag: 0.14.1
```
pchila, line 58:
```suggestion
              tag: 0.14.2
```

README.md
TannerGabriel: You are right. I just added a comment about that.

christian-kreuzberger-dtx (approved):
LGTM

PR #297, opened 2022-06-29
ci: use new keptn install action

pchila:
Switch to [action-install-keptn v2](https://github.com/keptn-sandbox/action-install-keptn/tree/v2.0.0).
Remove dependencies on Keptn CLI.


Raffy23 (approved):
LGTM

PR #309, opened 2022-07-07
fix: Handling of directories in the init container 

Raffy23:
## This PR
- fixes the way of how directories are handled in the init container


christian-kreuzberger-dtx (dismissed):
LGTM

christian-kreuzberger-dtx (approved):
LGTM

PR #314, opened 2022-07-11
feat: Add Keptn dev version to integration tests

TannerGabriel:
## This PR

- Fetches the latest Keptn release and prerelease
- Adds the versions to the Github action matrix
- Checks if the Keptn chart is available in the default Helm repository (`https://charts.keptn.sh`). If not, `https://charts-dev.keptn.sh` is used instead.

Integration Tests: https://github.com/keptn-contrib/job-executor-service/actions/runs/2649231941

### DIFF
--- a/.github/actions/github-script/fetch_keptn_versions.sh
+++ b/.github/actions/github-script/fetch_keptn_versions.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+# Fetch latest release and prerelease
+prerelease=$(curl -s https://api.github.com/repos/keptn/keptn/releases | jq -r 'map(select(.prerelease)) | first | .tag_name')
+release=$(curl -s https://api.github.com/repos/keptn/keptn/releases | jq -r 'map(select(.prerelease==false)) | first | .tag_name')
+
+matrix_config=""
+
+# Write variables as output
+echo "::set-output name=LATEST_RELEASE::$release"
+echo "::set-output name=LATEST_PRERELEASE::$prerelease"
--- a/.github/actions/github-script/fetch_keptn_versions.sh
+++ b/.github/actions/github-script/fetch_keptn_versions.sh
@@ -1,10 +1,8 @@
 #!/bin/bash
 
 # Fetch latest release and prerelease
-prerelease=$(curl -s https://api.github.com/repos/keptn/keptn/releases | jq -r 'map(select(.prerelease)) | first | .tag_name')
-release=$(curl -s https://api.github.com/repos/keptn/keptn/releases | jq -r 'map(select(.prerelease==false)) | first | .tag_name')
-
-matrix_config=""
+prerelease=$(curl -s https://api.github.com/repos/keptn/keptn/releases | jq -r 'map(select(.prerelease)) | sort_by(.tag_name)[-1].tag_name')
+release=$(curl --silent "https://api.github.com/repos/keptn/keptn/releases/latest" | jq -r '.tag_name')
 
 # Write variables as output
 echo "::set-output name=LATEST_RELEASE::$release"
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -11,13 +11,28 @@ defaults:
     shell: bash
 
 jobs:
+  prepare:
+    name: Prepare integration test run
+    runs-on: ubuntu-20.04
+    outputs:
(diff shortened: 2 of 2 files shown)


.github/actions/github-script/fetch_keptn_versions.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# Fetch latest release and prerelease
+prerelease=$(curl -s https://api.github.com/repos/keptn/keptn/releases | jq -r 'map(select(.prerelease)) | first | .tag_name')
+release=$(curl -s https://api.github.com/repos/keptn/keptn/releases | jq -r 'map(select(.prerelease==false)) | first | .tag_name')
```
Raffy23, line 5: Since the releases page does include the releases in the order they are published (please correct me if I'm wrong), I don't think that the first release is every time the latest. Imagine the following scenario: 
- Keptn releases 0.18.0
- A bug is discovered in 0.16.1 so they release a fix with the version 0.16.2

Since the fix was releases after 0.18.0, it will probably be the first one in the list and I don't think that we want to test this version as latest ...

.github/workflows/integration-tests.yaml
```diff
@@ -116,9 +131,27 @@ jobs:
...
+          echo "::set-output name=HELM_REPO::https://charts.keptn.sh"
+          else
+          echo "Using dev Helm repo"
+          echo "::set-output name=HELM_REPO::https://charts-dev.keptn.sh"
+          fi
+
       - name: Install Keptn
         id: install_keptn
-        uses: keptn-sandbox/action-install-keptn@v2.0.0
+        uses: keptn-sandbox/action-install-keptn@main
```
Raffy23, line 154: I think that we should pin a version here to avoid breaking things in the future if a new major version is released

TannerGabriel: Yes, we didn't have a new release for the Helm repository changes, but we will release a new version after fixing the Helm values for Keptn 0.17. After the release, I will define the version here.

.github/actions/github-script/fetch_keptn_versions.sh
```diff
@@ -0,0 +1,11 @@
...
+
+matrix_config=""
```
pchila, line 7: What do you need `matrix_config` for ?

.github/actions/github-script/fetch_keptn_versions.sh
TannerGabriel: After testing it a bit, it seems that the endpoint returns the items in the same order as on the Github releases page. We could use the `/latest` endpoint, which always gets the latest release of the repository, but the prerelease does not have such an endpoint. 

For fetching the prerelease, I believe the best solution would be to filter using the `tag_name`, which seems to work fine for the Keptn repository.

.github/workflows/integration-tests.yaml
```diff
@@ -116,9 +131,27 @@ jobs:
...
+        id: helm_repo_check
+        shell: bash
+        env:
+          KEPTN_VERSION: ${{ matrix.keptn-version }}
+        run: |
+          echo "Keptn Version: $KEPTN_VERSION"
+          helm repo add keptn https://charts.keptn.sh
+          helm_search=$(helm search repo keptn/keptn --version $KEPTN_VERSION)
+          
+          if [[ "$helm_search" != "No results found" ]]; then
```
pchila, line 144: Please do not use freetext output for testing, Try using `-o json` along with `jq` to test how many elements are returned.

TannerGabriel: Sounds like a more stable solution. I just changed it.

.github/workflows/integration-tests.yaml
```diff
@@ -11,13 +11,28 @@ defaults:
     shell: bash
 
 jobs:
+  prepare:
```
christian-kreuzberger-dtx, line 14: can we call this job "fetch-latest-keptn-versions" or alike?

christian-kreuzberger-dtx (commented):
I like the approach done here! Before we merge, let's addresses the comments, i.e., make sure that we have a new release of keptn/gh-automation.

.github/workflows/integration-tests.yaml
TannerGabriel: Sure just changed it.

.github/workflows/integration-tests.yaml
TannerGabriel: I just released a new version and added it here.

pchila (approved):
LGTM

PR #319, opened 2022-07-15
fix: Auto-detection of Keptn 0.17.0

Raffy23:
## This PR
- fixes the auto-detection for Keptn 0.17.0
- the defined namespace is now used even if auto-detection is enabled (to avoid the namespace lookup)


pchila (dismissed):
LGTM

christian-kreuzberger-dtx (approved):
LGTM

pchila (approved):

PR #327, opened 2022-07-20
chore: Update pipeline to be compatible with Keptn 0.17

christian-kreuzberger-dtx:
With PR #314 we are now testing against newer Keptn version (latest release and latest pre-release). However, it seems that we're not compatible.
image

This PR updates pipeline code to be compatible with Keptn 0.17 (note: this only proves that we're compatible with Keptn 0.17, we haven't put any effort into making it compatible, e.g., using a new distributor version).

I also slimmed down and updated the test matrix for Keptn versions, e.g.:
* Removed 0.13.x (we have plenty of 0.13 compatible releases that people can use)
* Removed 0.14.x (we have plenty of 0.14 compatible releases that people can use)
* Keep 0.15.1
* Update from 0.16.0 to 0.16.1

Integration Test Run: https://github.com/keptn-contrib/job-executor-service/actions/runs/2703312389

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


TannerGabriel (approved):
LGTM

PR #334, opened 2022-07-25
fix: Sending error logs to all registered jes instances in uniform

Raffy23:
## This PR
- changes sending error logs to Keptn uniform to all registered jes instances

### DIFF
--- a/pkg/keptn/errorlog_test.go
+++ b/pkg/keptn/errorlog_test.go
@@ -129,13 +129,15 @@ func TestErrorWhenMultipleMatchingRegistrationReturned(t *testing.T) {
 	).Times(1)
 
 	mockCloudEventSender := fake.NewMockCloudEventSender(ctrl)
+	mockCloudEventSender.EXPECT().SendCloudEvent(gomock.Any()).Times(2).Return(nil)
+
 	sut := NewErrorLogSender("foobar", uniformClient, mockCloudEventSender)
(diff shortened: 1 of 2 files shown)


pkg/keptn/errorlog_test.go
```diff
@@ -129,13 +129,15 @@ func TestErrorWhenMultipleMatchingRegistrationReturned(t *testing.T) {
...
 	sut := NewErrorLogSender("foobar", uniformClient, mockCloudEventSender)
 
 	newEvent := cloudevents.NewEvent()
-	err := sut.SendErrorLogEvent(&newEvent, errors.New("some error"))
+	newEvent.SetExtension("shkeptncontext", "034579438ufj-958340958")
 
-	assert.Error(t, err)
-	assert.ErrorContains(t, err, "found multiple uniform registrations with name foobar")
+	err := sut.SendErrorLogEvent(&newEvent, errors.New("some error"))
+	assert.NoError(t, err)
```
pchila, line 140: Name of test should be changed from TestErrorWhenMultipleMatchingRegistrationReturned to TestSendToAllRegistrationFound.

This can be done when the proper fix for the issue comes along

pchila (approved):
This is ok as a quick workaround but a proper fix should be implemented later where we filter on name, host and namespace to find the most specific registration for JES

Issue #145, opened 2021-12-30
Allow using a project-wide `job/config.yaml` (e.g., from main/master branch)


christian-kreuzberger-dtx:
Instead of adding a `job/config.yaml` for every service and stage, we should support adding the file once, e.g.

```bash
keptn add-resource --project=$PROJECT --resource=job-executor-config.yaml --resourceUri=job/config.yaml
```

IMO there are three possibilities for job/config.yaml to live:

* simple setup: project wide (**my proposed strategy**)
* complex setup: one job/config.yaml per service (**current strategy**)
* very complex setup: one job/config.yaml per service and stage (this would only be needed for customizations, which surely could also be done via separate config files for the tools used in job/config.yaml)

The **simple setup** has multiple benefits for me:

* shipyard.yaml and job/config.yaml are on the same branch and visually strongly connected
* less code duplication (my job/config.yaml looks the same for all services and stages) and therefore better maintenance

## Discussion

1: Should we use the simple setup?

2: If so, should we deprecate the complex setup, or should we support both?

3: If we support both, which of the two files is "stronger" in case of both files existing?

PR #338, opened 2022-08-03
feat: Add global and stage job configuration lookup

Raffy23:
## This PR
- Adds a way to set default configurations on project or stage level
- Implements #145

### DIFF
--- a/docs/FEATURES.md
+++ b/docs/FEATURES.md
@@ -3,6 +3,7 @@
 
 - [Features](#features)
   - [Getting started](#getting-started)
+  - [Job Configuration](#job-configuration)
   - [Specifying the working directory](#specifying-the-working-directory)
   - [Event Matching](#event-matching)
   - [Kubernetes Job](#kubernetes-job)
@@ -35,6 +36,29 @@
 
 To get started with job-executor-service, please follow the [Quickstart](../README.md#quickstart).
 
+### Job Configuration
+
+The job-executor-service can be configured with the `job/config.yaml` configuration file, which should be uploaded as
+resource to Keptn. The job-executor-service will scan the resources in the following order to determine which configuration
+should be used: 
+- **Service**: A resource that is stored on the service level will always used by the job executor service if available 
+- **Stage**: If no service resource is found, the stage resources are searched for a job configuration
+- **Project**: If no other resources are round the job executor will fallback to a project wide configuration.
+
+If the job executor service can't find a configuration file, it will respond with an error event, which can be viewed
+in the uniform page of the Keptn bridge.
+
+A typical job configuration contains usually one or more actions that will be triggered when a specific event is
+received: 
+```yaml
+apiVersion: v2
+actions:
+  - name: "Print files"
+    events:
+      - name: "sh.keptn.event.sample.triggered"
+    tasks:
+      - ...
+```
 
 ### Specifying the working directory
 
--- a/pkg/keptn/resource_service.go
+++ b/pkg/keptn/resource_service.go
@@ -8,6 +8,12 @@ import (
 	"net/url"
 )
 
+type ResourceHandler interface {
+	GetServiceResource(resource string, gitCommitID string) ([]byte, error)
+	GetStageResource(resource string, gitCommitID string) ([]byte, error)
+	GetProjectResource(resource string, gitCommitID string) ([]byte, error)
+}
+
 // V1ResourceHandler is a wrapper around the v1 ResourceHandler of the Keptn API to simplify the
 // getting of resources of a given event
 type V1ResourceHandler struct {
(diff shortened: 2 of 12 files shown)


docs/FEATURES.md
```diff
@@ -35,6 +36,29 @@
...
 To get started with job-executor-service, please follow the [Quickstart](../README.md#quickstart).
 
+### Job Configuration
+
+The job-executor-service can be configured with the `job/config.yaml` configuration file, which should be uploaded as
+resource to Keptn. The job-executor-service will scan the resources in the following order to determine which configuration
+should be used: 
+- **Service**: A resource that is stored on the service level will always used by the job executor service if available 
+- **Stage**: If no service resource is found, the stage resources are searched for a job configuration
+- **Project**: If no other resources are round the job executor will fallback to a project wide configuration.
```
TannerGabriel, line 46:
```suggestion
- **Project**: If no other resources are found the job executor will fallback to a project wide configuration.
```

docs/FEATURES.md
```diff
@@ -35,6 +36,29 @@
...
+
+If the job executor service can't find a configuration file, it will respond with an error event, which can be viewed
+in the uniform page of the Keptn bridge.
+
+A typical job configuration contains usually one or more actions that will be triggered when a specific event is
```
TannerGabriel, line 51:
```suggestion
A typical job configuration usually contains one or more actions that will be triggered when a specific event is
```

TannerGabriel (commented):
Just some typos I found looking over the PR.

christian-kreuzberger-dtx (dismissed):
I see some duplicated code between keptn/resource_service.go and keptn/config_service.go, but that is to be expected and not part of this PR to be fixed.
I will create a follow-up story for that.

Other then that, LGTM

pkg/keptn/resource_service.go
```diff
@@ -51,7 +52,47 @@ func (r V1ResourceHandler) GetResource(resource string, gitCommitID string) ([]b
...
+	resourceContent, err := r.ResourceHandler.GetResource(*scope, buildResourceHandlerV1Options(gitCommitID))
+	if err != nil {
+		return nil, fmt.Errorf("unable to get resouce from keptn: %w", err)
+	}
+
+	return []byte(resourceContent.ResourceContent), nil
+}
+
+// GetProjectResource returns the resource that was defined on project level
+func (r V1ResourceHandler) GetProjectResource(resource string, gitCommitID string) ([]byte, error) {
```
TannerGabriel, line 75: We could remove the `gitCommitID` parameter since it does not work as you mentioned. If not, we should probably document it in the function comment.

pkg/keptn/resource_service.go
```diff
@@ -51,7 +52,47 @@ func (r V1ResourceHandler) GetResource(resource string, gitCommitID string) ([]b
 		queryParam = api.AppendQuery(url.Values{})
 	}
 
-	resourceContent, err := r.ResourceHandler.GetResource(*scope, queryParam)
+	return queryParam
+}
+
+// GetServiceResource returns the contents of a resource for a given gitCommitID
+func (r V1ResourceHandler) GetServiceResource(resource string, gitCommitID string) ([]byte, error) {
```
TannerGabriel, line 59: Maybe we should document that an empty `gitCommitID` equals getting the resource without `gitCommitID`, or we could also make the parameter optional, e.g. (`...string`).

TannerGabriel (commented):
LGTM. Just some small improvement ideas that are optional.

pkg/keptn/resource_service.go
christian-kreuzberger-dtx: `gitCommitID` works fine on Keptn side, so it should stay in here.
though when WE use it as part of a task/sequence, the git commit id always refers to service/stage level, not project level. therefore we ommit the parameter in our calling code, but not in the actual implementation.

pkg/keptn/resource_service.go
christian-kreuzberger-dtx: I'd rather have it in here, same as keptn/go-utils does.
I would however document that when specifying an emtpy gitCommitID, Keptn will (most likely) just return the latest version of the file.

pkg/keptn/resource_service.go
TannerGabriel: Ok, I thought that we are always calling it with `gitCommitID=""`  because of that. If that is not the case, we can leave it as is.

pkg/keptn/resource_service.go
TannerGabriel: Ok sure

TannerGabriel (approved):
LGTM

christian-kreuzberger-dtx (approved):
LGTM

PR #351, opened 2022-08-29
feat!: Reimplement service with go-sdk

TannerGabriel:
## This PR

- Removes the distributor
- Reimplements the service using the `go-sdk` instead of `keptnutils` 
- Update `go-utils` to version `v0.18.0` 
- Removes the `useLocalFileSystem` flag and functionality

Note: The `TestJobExecutorServiceNoIngress` integration test is suppose to fail and will be removed in another PR


cmd/job-executor-service-initcontainer/main.go
```diff
@@ -51,6 +54,8 @@ type envConfig struct {
 	OAuthDiscovery string `envconfig:"OAUTH_DISCOVERY" required:"false"`
 	// The gitCommitId of the initial cloud event, for older Keptn instances this might be empty
 	GitCommitID string `envconfig:"GIT_COMMIT_ID"`
+	// The URL to the resource service
+	ConfigurationServiceURL string `envconfig:"CONFIGURATION_SERVICE" default:"resource-service:8080"`
```
christian-kreuzberger-dtx, line 58: This default value will never work, as it's usually not running in the same namespace as `keptn`.
I recommend setting required true.

cmd/job-executor-service-initcontainer/main.go
```diff
@@ -112,25 +118,20 @@ func main() {
...
-		Service:     env.Service,
-		GitCommitID: env.GitCommitID,
+	eventProps := &keptnv2.EventData{
+		Project: env.Project,
+		Stage:   env.Stage,
+		Service: env.Service,
 	}
 
-	configService := keptn.NewConfigService(useLocalFileSystem, eventProps, keptnAPI.Resources())
+	resourceService := keptn_interface.NewV1ResourceHandler(*eventProps, keptnAPI.Resources())
```
christian-kreuzberger-dtx, line 128: This is missing `gitCommitID` - V1 resource handler probably doesn't support it?

cmd/job-executor-service-initcontainer/main.go
christian-kreuzberger-dtx: Actually, is this needed? Doesn't look like `ConfigurationServiceURL` is used in main.go

cmd/job-executor-service/main.go
```diff
@@ -223,43 +192,60 @@ func _main(args []string, env envConfig) int {
...
+			eventWildcard,
+			NewEventHandler(imageFilterList),
+			jobEventFilter),
+		sdk.WithLogger(logrus.New()),
+	).Start())
+}
 
-	if err != nil {
-		log.Fatalf("failed to create client, %v", err)
+func jobEventFilter(keptnHandle sdk.IKeptn, event sdk.KeptnEvent) bool {
```
christian-kreuzberger-dtx, line 210: please add a comment what this function is doing (e.g., "Checking if a job/config exists and whether it contains the event type")

cmd/job-executor-service/main.go
```diff
@@ -223,43 +192,60 @@ func _main(args []string, env envConfig) int {
...
-	p, err := cloudevents.NewHTTP(
-		cloudevents.WithPath(env.Path),
-		cloudevents.WithListener(
-			listener,
-		),
-	)
+	log.Fatal(sdk.NewKeptn(
+		serviceName,
+		sdk.WithTaskHandler(
+			eventWildcard,
```
christian-kreuzberger-dtx, line 203: Please add a comment like "handle all events and filter them later in jobEventFilter"

pkg/config/reader.go
```diff
@@ -6,6 +6,8 @@ import (
 	"log"
 )
 
+// TODO: Escape /
```
christian-kreuzberger-dtx, line 9: Is this still a todo? Or did we fix this in the code

cmd/job-executor-service/main.go
```diff
@@ -223,43 +192,60 @@ func _main(args []string, env envConfig) int {
...
+		Keptn: keptn_interface.NewV1ResourceHandler(*data, keptnHandle.APIV2().Resources()),
 	}
 
-	imageFilterList, err := utils.BuildImageAllowList(env.AllowedImageList)
+	// Check if the job configuration can be found
+	configuration, _, err := jcr.GetJobConfig(event.GitCommitID)
+
 	if err != nil {
-		log.Fatalf("failed to generate the allowlist, %v", err)
+		keptnHandle.Logger().Infof("could not retrieve config for job-executor-service: %e", err)
```
christian-kreuzberger-dtx, line 227: We should send an error log event, which was previously done by the eventhandler itself, e.g.:
```
errorLogErr := eh.ErrorSender.SendErrorLogEvent(
			eh.Keptn.CloudEvent, fmt.Errorf(
				"could not retrieve config for job-executor-service: %w", err,
			),
		)
```

README.md
```diff
@@ -35,7 +35,7 @@ running in the Keptn ecosystem:
 |    0.16.x     |                               keptncontrib/job-executor-service:0.2.3                                |       v2       |
 |    0.16.x     |                               keptncontrib/job-executor-service:0.2.4                                |       v2       |
 |    0.17.x     |                               keptncontrib/job-executor-service:0.2.5                                |       v2       |
-
+|    0.18.x     |                               keptncontrib/job-executor-service:0.2.6                                |       v2       |
```
Raffy23, line 38: I assume that we are removing version 0.16.1 and 0.17.0 from the test matrix because this release isn't compatible anymore. If so this should also be reflected in the version number and this PR should be marked as a breaking change.

cmd/job-executor-service-initcontainer/main.go
```diff
@@ -63,6 +68,7 @@ func main() {
 	fs := afero.NewOsFs()
 
 	baseURL, err := url.Parse(env.KeptnAPIURL)
+	log.Printf("BaseURL casting: %s", baseURL)
```
Raffy23, line 71: Is this `log.Printf` a thing left from debugging code, or is it intended for the output? If so I would prefer to have a different message, like "Keptn API URL: %s"

cmd/job-executor-service-initcontainer/main.go
```diff
@@ -112,25 +118,20 @@ func main() {
...
-		log.Println("env=local: Running with local filesystem to fetch resources")
-		useLocalFileSystem = true
-	}
-
 	// re-create the event from job-executor-service
-	eventProps := keptn.EventProperties{
-		Project:     env.Project,
-		Stage:       env.Stage,
-		Service:     env.Service,
-		GitCommitID: env.GitCommitID,
```
Raffy23, line 128: Why was the GitCommitID removed here? If we use the resource-service we should always try to fetch the resource with the provided commit id

pkg/eventhandler/eventhandlers_test.go
```diff
@@ -142,7 +155,7 @@ func (mei *matcherErrorIs) String() string {
 	return fmt.Sprintf("%#v", mei.targetErr)
 }
 
-func TestErrorGettingJobConfig(t *testing.T) {
+/*func TestErrorGettingJobConfig(t *testing.T) {
```
Raffy23, line 158: Is there any reason why this test is commented out?

pkg/file/file_test.go
```diff
@@ -44,33 +38,38 @@ const yamlFile = `
...
 
 	mockCtrl := gomock.NewController(t)
 	defer mockCtrl.Finish()
 
-	return keptnfake.NewMockConfigService(mockCtrl)
+	return keptnfake.NewMockResourceHandler(mockCtrl)
 }
 
-func TestMountFiles(t *testing.T) {
+/*func TestMountFiles(t *testing.T) {
```
Raffy23, line 49: I think that we should still test the mounting of the files logic

pkg/k8sutils/job.go
```diff
@@ -329,6 +329,10 @@ func (k8s *K8sImpl) CreateK8sJob(
 									Name:  "GIT_COMMIT_ID",
 									Value: jobDetails.GitCommitID,
 								},
+								{
+									Name:  "CONFIGURATION_SERVICE",
```
Raffy23, line 333: I think that variable is unused in the init container and if not it should probably be renamed to `RESOURCE_SERVICE`

pkg/keptn/resource_service.go
```diff
@@ -61,10 +88,16 @@ func (r V1ResourceHandler) GetServiceResource(resource string, gitCommitID strin
 	scope.Service(r.Event.Service)
 	scope.Project(r.Event.Project)
 	scope.Stage(r.Event.Stage)
-	scope.Resource(resource)
+	scope.Resource(url.QueryEscape(resource))
+
+	log.Printf("Resource: %s", scope.GetResourcePath())
```
Raffy23, line 93: I would prefer to not have logging output here, if you really want to preserve this log line it should be more descriptive like `"Fetching Resource from path %s"`

pkg/keptn/resource_service.go
```diff
@@ -61,10 +88,16 @@ func (r V1ResourceHandler) GetServiceResource(resource string, gitCommitID strin
...
 
-	resourceContent, err := r.ResourceHandler.GetResource(*scope, buildResourceHandlerV1Options(gitCommitID))
+	options := api.ResourcesGetResourceOptions{URIOptions: []api.URIOption{
+		buildResourceHandlerV1Options(gitCommitID),
+	}}
+	resourceContent, err := r.ResourceAPI.GetResource(context.Background(), *scope, options)
 	if err != nil {
+		log.Printf("unable to get resouce from keptn: %e", err)
```
Raffy23, line 100: Is there any reason why the error will be printed into the log and returned?

pkg/keptn/resource_service.go
```diff
@@ -75,10 +108,15 @@ func (r V1ResourceHandler) GetServiceResource(resource string, gitCommitID strin
...
+	scope.Resource(url.QueryEscape(resource))
 
-	resourceContent, err := r.ResourceHandler.GetResource(*scope, buildResourceHandlerV1Options(gitCommitID))
+	log.Printf("Resource: %s", scope.GetResourcePath())
+	options := api.ResourcesGetResourceOptions{URIOptions: []api.URIOption{
+		buildResourceHandlerV1Options(gitCommitID),
+	}}
+	resourceContent, err := r.ResourceAPI.GetResource(context.Background(), *scope, options)
 	if err != nil {
+		log.Printf("unable to get resouce from keptn: %e", err)
```
Raffy23, line 119: Same remarks as above

pkg/keptn/resource_service.go
```diff
@@ -90,12 +128,71 @@ func (r V1ResourceHandler) GetStageResource(resource string, gitCommitID string)
...
```
Raffy23, line 139: Same remarks as above

cmd/job-executor-service-initcontainer/main.go
TannerGabriel: No, this is not needed in the `init-container` and has been removed.

cmd/job-executor-service-initcontainer/main.go
TannerGabriel: The `gitCommitID` that was passed to the constructor was not used. You have to set it whenever calling a function e.g. `GetJobConfig(gitCommitID string)`. Still, I added the `gitCommitID` to the `mountFiles` method, so this should be fixed.

pkg/config/reader.go
TannerGabriel: No just removed it

README.md
TannerGabriel: You are right. Since we are using the new `resource-service` endpoint, it is no longer compatible with older versions. I have marked the PR as breaking.

cmd/job-executor-service-initcontainer/main.go
TannerGabriel: Yes, that was a debug statement and should be removed.

cmd/job-executor-service-initcontainer/main.go
TannerGabriel: The `&keptnv2.EventData` does not have a `GitCommitID` field and the value was also not used in the constructor of the `ResourceHandler`. [This comment](https://github.com/keptn-contrib/job-executor-service/pull/351#discussion_r958420056) should provide more information.

pkg/file/file_test.go
TannerGabriel: Sorry, these tests have been commented out while testing. I just added them again with the new mocks and dependencies.

pkg/k8sutils/job.go
TannerGabriel: Has already been removed ([see](https://github.com/keptn-contrib/job-executor-service/pull/351#discussion_r958417502))

cmd/job-executor-service/main.go
TannerGabriel: Let's do this in a separate PR since this one is already quite large.

pkg/eventhandler/eventhandlers.go
```diff
@@ -304,7 +286,7 @@ func sendJobFailedEvent(myKeptn *keptnv2.Keptn, jobName string, serviceName stri
 	}
 }
 
-func sendTaskFinishedEvent(myKeptn *keptnv2.Keptn, serviceName string, jobLogs []jobLogs, data dataForFinishedEvent) {
+func getTaskFinishedEvent(event sdk.KeptnEvent, receivedEventData keptn.EventProperties, jobLogs []jobLogs, data dataForFinishedEvent) interface{} {
```
Raffy23, line 289: I don't think that the name of this function is suitable, as it computes either a `keptnv2.TestFinishedEventData` or `&keptnv2.EventData` event and returns an `interface{}`, maybe this function should be refactored a little bit to make use of some type in the return value.

pkg/eventhandler/eventhandlers_test.go
TannerGabriel: It seems like I forgot to rewrite the tests to use the `go-sdk`. It should be added now.

pkg/eventhandler/eventhandlers.go
TannerGabriel: I'm not sure if this is needed since we only use the function in combination with the `go-sdk`, which also uses `interface{}`, but I'm open to suggestions.

Raffy23: If that's the case, please add this as comment to the function header that explains the need for using the `interface{}` type. At least that way it's properly documented and avoid confusion in the future.

Raffy23 (approved):
LGTM

PR #353, opened 2022-08-31
chore: Delete NoIngress integration test

TannerGabriel:
## This PR
- Removes the `noingress_test.go` file since it is no longer needed after the deletion of the distributor

Integration test run: https://github.com/keptn-contrib/job-executor-service/actions/runs/2963548495


Raffy23 (approved):
Since there is the need for the health-check the test can't work as expected.
LGTM!

PR #354, opened 2022-08-31
feat: Send error cloud event

TannerGabriel:
## This PR

- Reimplement `ErrorLogSender` with go-sdk
- Send error event when `job/config.yaml` cannot be found


pkg/keptn/errorlog.go
```diff
@@ -96,34 +102,20 @@ func (els *ErrorLogSender) SendErrorLogEvent(initialCloudEvent *cloudevents.Even
 	return fmt.Errorf("no registration found with name %s", els.integrationName)
 }
 
-func createErrorLogCloudEvent(
-	integrationID string, initialEvent *cloudevents.Event, err error,
-) (cloudevents.Event, error) {
-	errorData := ErrorData{
+func createErrorLog(
+	integrationID string, initialEvent *sdk.KeptnEvent, err error,
+) (models.LogEntry, error) {
```
Raffy23, line 107: It seems like the function never returns an error

TannerGabriel: It seems like I forgot to change the return types. I fixed it now.

Raffy23 (approved):
LGTM

PR #356, opened 2022-09-06
chore: Update dependencies

Raffy23:
## This PR
- updates all dependencies


TannerGabriel (commented):
We should probably also update the go version in the Github actions scripts (CI, Integration tests, etc.).

Raffy23:
I will double check later, but I think all GitHub action should get the Go version from the go mod file

TannerGabriel:
> I will double check later, but I think all GitHub action should get the Go version from the go mod file

If that is the case, we can most likely remove this [environment variable](https://github.com/keptn-contrib/job-executor-service/blob/main/.github/workflows/integration-tests.yaml#L39).

christian-kreuzberger-dtx:
Can we please separate the go 1.18 update into another PR with a better title? :) Would be great for the changelog.

Dockerfile
```diff
@@ -1,7 +1,7 @@
 # Use the offical Golang image to create a build artifact.
 # This is based on Debian and sets the GOPATH to /go.
 # https://hub.docker.com/_/golang
-FROM golang:1.17.5-alpine as builder
+FROM golang:1.18-alpine as builder
```
christian-kreuzberger-dtx, line 4: Please use a more specific go version, e.g., 1.18.3 (as that's the last official supported version of Dynatrace oneagent https://www.dynatrace.com/support/help/technology-support/application-software/go/support/supported-go-versions )

TannerGabriel (approved):
LGTM

Raffy23:
PR depends on #357

TannerGabriel (approved):
LGTM

PR #357, opened 2022-09-06
chore: Update to Go 1.18

Raffy23:
## This PR 
- updates JES to Go 1.18
- fixes reviewdog action

### DIFF
--- a/.github/actions/unit-tests/action.yml
+++ b/.github/actions/unit-tests/action.yml
@@ -2,7 +2,7 @@ name: "Unit Tests"
 description: "Run unit tests using go"
 inputs:
   GO_VERSION:
-    default: "1.17"
+    default: "1.18"
 env:
   GO111MODULE: "on"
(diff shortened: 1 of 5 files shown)


.github/actions/unit-tests/action.yml
```diff
@@ -2,7 +2,7 @@ name: "Unit Tests"
 description: "Run unit tests using go"
 inputs:
   GO_VERSION:
-    default: "1.17"
+    default: "1.18"
```
TannerGabriel, line 5: This is unused, but it seems like you are removing it in #356. Is that correct?

TannerGabriel (approved):
LGTM

.github/actions/unit-tests/action.yml
Raffy23: Yes, this should be unused, but I used this PR to just update the only Go version everywhere and make changes to the GitHub action in the other ones to keep this one small.

